PR #2, opened 2012-10-03
code style fixes, couldn't test due to a lack of a working module that

willtp87:
smoke tested

### DIFF
--- a/modules/islandora_scholar_embargo/embargo.tab.inc
+++ b/modules/islandora_scholar_embargo/embargo.tab.inc
@@ -17,7 +17,11 @@ class EmbargoView {
    */
   public static function get_tabs($pid) {
     $tab_set = array();
-    if (user_access('ingest new fedora objects') || $is_owner) {
+    /*
+     * @todo implement $is_owner access
+     * if (user_access('ingest new fedora objects') || $is_owner) {
+     */
+    if (user_access('ingest new fedora objects')) {
       $tab_set['embargo'] = array(
         '#title' => t('Embargo'),
         '#type' => 'tabpage',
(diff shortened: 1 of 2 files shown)


modules/islandora_scholar_embargo/islandora_scholar_embargo.module
```diff
@@ -1,12 +1,30 @@
...
+/**
+ * @file
+ *   This is the main module file for islandora_scholar_embargo.
+ *   It supports embargo management in Islandora Scholar.
+ */
 
-//Doesn't make sense to use either of the Islandora RELS-EXT or RELS-INT
-//  namespaces, as our predicates can occur in both.
+// Doesn't make sense to use either of the Islandora RELS-EXT or RELS-INT
+// Namespaces, as our predicates can occur in both.
```
adam-vessey: Comment here is kinda broken...  This is more of a multi-line comment, not multiple distinct comments.

modules/islandora_scholar_embargo/embargo.tab.inc
```diff
@@ -147,26 +166,35 @@ WHERE {{
...
       list($e_pid, $e_dsid) = explode('/', $e_pid);
     }
 
-    //XXX:  Might get into trouble if we somehow have multiple embargo
-    //  statements for a single datastream, due to how we remove them from the
-    //  XACML.
-    //Do not remove statements which match our parameters.
+    /* XXX:  Might get into trouble if we somehow have multiple embargo
+      statements for a single datastream, due to how we remove them from the
+      XACML.*/
```
adam-vessey, line 180: According to http://drupal.org/node/1354#inline, C++ style comments ("//") are preferred.

willtp87, line 180: Preferred, not demanded.  Given how code sniffer doesn't like the indentation convention previously noted on the multiple line  // comment of the original code I think this is an acceptable trade to indicate that this is a single multiline comment.

PR #7, opened 2012-12-11
Fix Solr RSS feeds 535 second try

philsogaDGI:


includes/SolrResults.inc
```diff
@@ -140,20 +142,22 @@ function template_preprocess_islandora_scholar_search_item(array &$variables) {
...
   $total = (int) $results->response->numFound;
   $start = (int) $query_processor->solrStart + 1;
   $end = min(($query_processor->solrLimit + $query_processor->solrStart), $total);
-  $title = t('Search Results (%start - %end of %total)', array('%start' => $start, '%end' => $end, '%total' => $total));
+  //$rss_link = $this->addSecondaries($query_processor);
+  $title = t('Search Results (%start - %end of %total)' , array('%start' => $start, '%end' => $end, '%total' => $total));
   $table = array(
     '#header' => array($title),
     '#theme' => THEME_SCHOLAR_SEARCH_TABLE,
+    '#value' => $secondary_display,
```
DannyJoris, line 156: I don't think this is the best place for this. If it needs to be in the form array, it could be in it's own form element with markup.
http://api.drupal.org/api/drupal/developer!topics!forms_api_reference.html/6#markup

DannyJoris, line 156: well, actually, it already is in a form array as '#value'. I guess it's ok.

PR #8, opened 2013-02-08
first commit of Embargo module in 7

ajstanley:


modules/islandora_scholar_embargo/islandora_scholar_embargo.module
```diff
@@ -69,27 +103,26 @@ function islandora_scholar_embargo_mail_tokens($params, $language) {
...
 function islandora_scholar_embargo_mail($key, &$message, $params) {
-  $language = $message['language'];
-  $variables = user_mail_tokens($params['account'], $language);
-  $variables += islandora_scholar_mail_tokens($params, $language);
+  global $language;
+  $variables = islandora_scholar_embargo_mail_tokens($params, 'English');
 
   switch ($key) {
     case 'pre_expiry':
-      $message['subject'] = t('Notification from !site', $variables, $language->language);
```
adam-vessey, line 78: Need to pass the language parameter to t(), as the message should be able to be translated into the language of the recipient. This change would make it be the language of the sender.

modules/islandora_scholar_embargo/islandora_scholar_embargo.module
```diff
@@ -102,96 +135,89 @@ function islandora_scholar_embargo_mail($key, &$message, $params) {
...
-    $results = ObjectHelper::performSparqlQuery(<<<EOQ
+  // lift expired embargos
+  $query = <<<EOQ
 PREFIX is: <$ise_uri>
 SELECT ?expired ?date
 FROM <#ri>
 WHERE {
   ?expired is:$eu ?date .
-  FILTER(sameTerm(dataType(?date), xsd:dateTime) && ?date <= $now^^xsd:dateTime) .
+  FILTER(?date < '$tomorrow'^^xsd:dateTime && ?date > '$yesterday'^^xsd:dateTime) .
```
adam-vessey, line 204: Any reason just to filter to just the current day? May be a bad assumption, that cron runs daily.

PR #11, opened 2013-02-11
made ingest stateful

ajstanley:


modules/islandora_scholar_embargo/includes/embargo.inc
```diff
@@ -15,48 +15,47 @@
...
-  /* Now we've tried filling $result,
-   * if theres nothing in it, then there's no embargo. */
-  if (!empty($result)) {
-    $embargo_stat = $result[0]['e']['value'];
-    if ($embargo_stat !== "indefinite") {
-      $embargo_date = $embargo_stat;
-      $embargo_stat = "date";
+    else {
+      $date = new dateTime($embargo_date);
+      $message = $date->format('M d, Y ');
```
adam-vessey, line 54: Should probably use Drupal's format_date() (http://api.drupal.org/api/drupal/includes!common.inc/function/format_date/7), so it uses whatever format is configured for the site.

PR #27, opened 2013-03-10
modified xpaths, updated the admin form, fixed and added elements

dmoses:
After working with the DOI importer, here are some bits that I thought
should be updated. In citeproc/includes/converter.inc I changed the
xpaths for //originInfo/date*. I don't think that will break anything
and may catch edge cases. I changed some text on the admin form and
provided examples. In doi_importer I fixed an element and added genre
and dateIssued. Tested on my local VM.


jonathangreen:
@jordandukart Can you test this out and merge if its all working?

PR #28, opened 2013-03-12
Fix to make authors display in our citation table correctly.

jordandukart:


jordandukart:
Alan had a look at this just doesn't not have permission to merge to DGI.

PR #51, opened 2013-07-24
Myriad of bug fixes and improvements.

jordandukart:
- UI Improvements
- RELS-INT trickery to be able to embargo objects that do not currently have the restricted datastream presently on it.
- Bug fixes.
- Remove "Object Policy" tab if the item is currently embargoed.


modules/islandora_scholar_embargo/includes/embargo.inc
```diff
@@ -22,109 +22,139 @@ function islandora_scholar_embargo_form(array $form, array &$form_state, $item)
   $pid = $item->id;
   $label = $item->label;
```
philsogaDGI: Doesn't look like $label is being used anywhere.

PR #53, opened 2013-07-29
 Google Scholar Search

philsogaDGI:


islandora_scholar.module
```diff
@@ -156,6 +156,28 @@ function islandora_scholar_ir_citationCModel_islandora_view_object($object) {
     $display['citation']['#markup'] = CitationView::getCitation($object['MODS']->content);
   }
 
+  // Google Scholar Search.
```
jordandukart: All this logic should shift down into the variable get condition as there's no point to grab these things if we aren't going to be using them.

islandora_scholar.module
```diff
@@ -156,6 +156,28 @@ function islandora_scholar_ir_citationCModel_islandora_view_object($object) {
...
+  $mods_xml = simplexml_load_string($object['MODS']->content);
```
jordandukart: Should also check that the MODS datastream exists before we attempt to grab the content. Likely your conditional would be `if (variable_get('islandora_scholar_google_scholar_search', FALSE) && isset($object['MODS']))`

islandora_scholar.module
```diff
@@ -156,6 +156,28 @@ function islandora_scholar_ir_citationCModel_islandora_view_object($object) {
...
+  $mods_xml->registerXPathNamespace('mods', 'http://www.loc.gov/mods/v3');
+
+  // Default search term is the title of the citation.
+  $title = $mods_xml->xpath('//mods:title');
+  $search_term = (string) reset($title);
+
+  // Search for doi for search term.  If exists use it instead of default.
```
jordandukart: Perhaps make the DOI XPATH configurable? For example in Berkeley's current form their ingest form creates DOIs using the MODS element: `<mods:identifier type="accession">111112</mods:identifier>`.

islandora_scholar.module
```diff
@@ -156,6 +156,28 @@ function islandora_scholar_ir_citationCModel_islandora_view_object($object) {
...
+  $doi = $mods_xml->xpath('//mods:identifier[@type="doi"]');
+  if ($doi) {
+    $search_term = (string) reset($doi);
+  }
+
+  if (variable_get('islandora_scholar_google_scholar_search', FALSE)) {
+    $display['google_scholar_search'] = array(
+      '#type' => 'item',
+      '#markup' => l(t('Search for this publication on Google Scholar'), "http://scholar.google.ca/scholar?q=$search_term"),
```
jordandukart: For more precise searching I think placing "s around the search query (whether they be url encoded or not) is probably beneficial such that we narrow down our search results.

islandora_scholar.module
```diff
@@ -133,7 +133,7 @@ function islandora_scholar_islandora_required_objects(IslandoraTuque $connection
 function islandora_scholar_ir_citationCModel_islandora_view_object($object) {
   module_load_include('inc', 'islandora_scholar', 'includes/citation.tab');
   module_load_include('inc', 'islandora_scholar', 'includes/coins');
-  $i = 0;
+  $i = 1;
```
jordandukart, line 136: Any specific reason as to why this variable changed?

includes/admin.form.inc
```diff
@@ -50,6 +56,16 @@ function islandora_scholar_admin_form() {
...
     '#description' => t('Normally, we would have all calls go through Drupal get the relevant PDF. This option tries to make the call go directly to your associated Fedora instance.  (Note:  This will not work if you use "localhost" or a non-public IP address in the Islandora\'s base Fedora URL.)'),
   ));
+  $form['google_scholar_search'] += _islandora_scholar_add_variable('islandora_scholar_google_scholar_search', FALSE, array(
+    '#type' => 'checkbox',
+    '#title' => t('Google Scholar Search.'),
+    '#description' => t('Enable Searching in Google Scholar.'),
+  ));
+  $form['google_scholar_search'] += _islandora_scholar_add_variable('islandora_scholar_google_scholar_search_field', "doi", array(
+    '#type' => 'textfield',
+    '#title' => t('Google Scholar Search field.'),
```
jordandukart: Period in title.

includes/admin.form.inc
```diff
@@ -50,6 +56,16 @@ function islandora_scholar_admin_form() {
...
+    '#description' => t('The MODS field you want to use to search.  Default is doi'),
```
jordandukart: Perhaps we should modify this to take an XPATH value or explain that is a mods:identifier[@type="thiswherethevariableis"].

islandora_scholar.module
```diff
@@ -156,6 +156,30 @@ function islandora_scholar_ir_citationCModel_islandora_view_object($object) {
...
+    $identifier = variable_get('islandora_scholar_google_scholar_search_field', FALSE);
+
+    $doi = $mods_xml->xpath("//mods:identifier[@type=\"$identifier\"]");
+    if ($doi) {
+      $search_term = (string) reset($doi);
+    }
+
+    $display['google_scholar_search'] = array(
+      '#type' => 'item',
+      '#markup' => l(t('Search for this publication on Google Scholar'), "http://scholar.google.ca/scholar?q=\"$search_term\""),
```
jordandukart, line 177: If we have an identifier we would want that prefixed in the search query such that if we have a DOI of "doi:1112" we search for "doi:1112" as opposed to just 1112.

islandora_scholar.module
```diff
@@ -156,6 +156,29 @@ function islandora_scholar_ir_citationCModel_islandora_view_object($object) {
...
+    $mods_xml->registerXPathNamespace('mods', 'http://www.loc.gov/mods/v3');
+
+    // Search for doi for search term.  If exists use it instead of default.
+    $doi = $mods_xml->xpath("" . variable_get('islandora_scholar_google_scholar_primary_search_xpath', ""));
+    if ($doi) {
+      $search_term = (string) reset($doi);
+    }
+    else {
+      // Default search term is the title of the citation.
+      $title = $mods_xml->xpath("" . variable_get('islandora_scholar_google_scholar_default_search_xpath', ""));
```
jordandukart: Should define the default for the variable get.

islandora_scholar.module
```diff
@@ -156,6 +156,29 @@ function islandora_scholar_ir_citationCModel_islandora_view_object($object) {
     $display['citation']['#markup'] = CitationView::getCitation($object['MODS']->content);
   }
 
+  if (variable_get('islandora_scholar_google_scholar_search', FALSE) && isset($object['MODS'])) {
+    // Google Scholar Search.
+    $mods_xml = simplexml_load_string($object['MODS']->content);
+    $mods_xml->registerXPathNamespace('mods', 'http://www.loc.gov/mods/v3');
+
+    // Search for doi for search term.  If exists use it instead of default.
+    $doi = $mods_xml->xpath("" . variable_get('islandora_scholar_google_scholar_primary_search_xpath', ""));
```
jordandukart: Should define the default for the variable get. Also this variable could probably be renamed to something else and have the comment updated.

PR #57, opened 2013-08-19
Removed what I think is an SVN artifact from three module .info files

ksclarke:
Three Islandora_Scholar .info files have $Format:%h$ in the version field (while the other .info files do not).  I think these should be removed and the version for these modules should just be 7.x-dev(?)


jonathangreen:
:thumbsup: thanks kevin!

PR #59, opened 2013-08-30
changed query to get pdf embargoed objects

philsogaDGI:


jordandukart:
Included in https://github.com/discoverygarden/islandora_scholar/pull/60/

PR #60, opened 2013-09-03
7.x manage embargo

jordandukart:


modules/islandora_scholar_embargo/includes/embargo_manage.inc
```diff
@@ -29,30 +29,49 @@ function islandora_scholar_embargo_manage_form(array $form, array &$form_state)
...
+    $pid = $result['obj']['value'];
+    $object_parts = explode('/', $pid);
+
+    // When we explode if there's more than 1 element it constitutes that we
+    // embargoed something other than the object itself.
+    if (count($object_parts) == 1) {
+      $type = t('Object');
+    }
+    else {
+      $type = $object_parts[1];
```
adam-vessey: Kinda nitpicking, but why not do `list($pid, $type) = $object_parts;`. Hurray, being concise! :P

PR #67, opened 2013-10-25
added scholarbatchimporter to override getIdentifier and check for confi...

ppound:
...gured namespaces

### DIFF
--- a/includes/importer.inc
+++ b/includes/importer.inc
@@ -1,0 +1,60 @@
+<?php
+/**
+ * @file
+ * Defines ScholarBatchImporter.
+ */
+
+abstract class ScholarBatchImporter extends IslandoraBatchImporter {
+
+  /**
+   * Get an identifier.
+   *
+   * @param IslandoraTuque $tuque
+   *   a tuque object
+   * @param string $namespace
+   *   a default namespace.  We use this namespace if we can't find a Collection
+   * Policy or the collection solution pack is not installed.
+   *
+   * @return string
+   *   An identifier for the repo.
+   */
+  protected function getIdentifier(IslandoraTuque $tuque, $namespace) {
+    static $cm_namespace;
+    if (!isset($cm_namespace)) {
+      $cm_namespace = $this->getNamespaceFromCollectionPolicy($namespace);
+    }
+    return parent::getIdentifier($tuque, $cm_namespace);
+  }
+
+  /**
+   * Returns the namespace configured in the collection policy if it exists.
+   *
+   * If the namespace is not defined for the ir:citationCModel it will return
+   * the default namespace provided.
+   *
+   * @param string $default_namespace
+   *   will return the default namespace if it can't find one in the collection
+   * policy or the collection solution pack is not installed.
+   *
+   * @return string
+   *   the namespace to use
+   */
+  protected function getNamespaceFromCollectionPolicy($default_namespace) {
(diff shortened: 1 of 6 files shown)


ppound:
forgot to put the jira ticket the comments https://jira.duraspace.org/browse/ISLANDORA-914

includes/importer.inc
```diff
@@ -0,0 +1,60 @@
...
+    if (!module_exists('islandora_basic_collection')) {
+      return $default_namespace;
+    }
+    $parent = islandora_object_load($this->parentPid);
+    if (isset($parent['COLLECTION_POLICY'])) {
+      module_load_include('inc', 'islandora_basic_collection', 'collection_policy');
+      $policy_stream = $parent['COLLECTION_POLICY'];
+      $policy = new CollectionPolicy($policy_stream->content);
+
+      foreach ($policy->getContentModels() as $pid => $info) {
```
adam-vessey: No need to iterate here... We know we're only concerned with `ir:citationCModel`.

includes/importer.inc
```diff
@@ -0,0 +1,60 @@
...
+   * @return string
+   *   the namespace to use
+   */
+  protected function getNamespaceFromCollectionPolicy($default_namespace) {
+    if (!module_exists('islandora_basic_collection')) {
+      return $default_namespace;
+    }
+    $parent = islandora_object_load($this->parentPid);
+    if (isset($parent['COLLECTION_POLICY'])) {
+      module_load_include('inc', 'islandora_basic_collection', 'collection_policy');
```
adam-vessey: No need to manually load this include... The `CollectionPolicy` class is already available via autoloading: https://github.com/Islandora/islandora_solution_pack_collection/blob/7.x/islandora_basic_collection.info#L10

includes/importer.inc
```diff
@@ -0,0 +1,60 @@
...
+   *   a tuque object
+   * @param string $namespace
+   *   a default namespace.  We use this namespace if we can't find a Collection
+   * Policy or the collection solution pack is not installed.
+   *
+   * @return string
+   *   An identifier for the repo.
+   */
+  protected function getIdentifier(IslandoraTuque $tuque, $namespace) {
+    static $cm_namespace;
```
adam-vessey, line 17: Should probably use `drupal_static()`.

adam-vessey, line 17: Eh... I could live without it, though.

includes/importer.inc
```diff
@@ -0,0 +1,60 @@
...
+
+      foreach ($policy->getContentModels() as $pid => $info) {
+        if ($pid == 'ir:citationCModel' && isset($info['namespace'])) {
+          return $info['namespace'];
+        }
+      }
+    }
+    return $default_namespace;
+  }
+}
```
adam-vessey: Missing a newline.

includes/importer.inc
```diff
@@ -0,0 +1,60 @@
+<?php
+/**
+ * @file
+ * Defines ScholarBatchImporter.
+ */
+
+abstract class ScholarBatchImporter extends IslandoraBatchImporter {
+
+  /**
+   * Get an identifier.
```
adam-vessey: This comment should be minimal (as in, the entire doxygen comment), as it's overriding an already-existing method...  Should probably just update the ancestor's comment, reference it here, and indicate what's happening differently.

PR #71, opened 2013-11-20
7.x pdf changes

jordandukart:


xml/DS-COMPOSITE-MODEL.xml
```diff
@@ -14,4 +14,13 @@
   <dsTypeModel ID="PDF" optional="true">
     <form MIME="application/pdf"/>
   </dsTypeModel>
+  <dsTypeModel ID="FULL_TEXT">
+    <form MIME="text/plain"></form>
+  </dsTypeModel>
+  <dsTypeModel ID="TN">
+    <form MIME="image/jpeg"></form>
+  </dsTypeModel>
+  <dsTypeModel ID="PREVIEW">
```
adam-vessey: These three derived datastreams should also be optional.

islandora_scholar.drush.inc
```diff
@@ -0,0 +1,70 @@
...
+    'examples' => array(
+      'drush -u 1 islandora-scholar-update-citations' => dt('Updating existing citations with PDF derivatives.'),
+    ),
+    'bootstrap' => DRUSH_BOOTSTRAP_DRUPAL_LOGIN,
+  );
+  return $commands;
+}
+
+/**
+ * Command callback to install required objects.
```
adam-vessey: Copy/pasta?

islandora_scholar.drush.inc
```diff
@@ -0,0 +1,70 @@
...
+      }
+    }
+    if ($success) {
+      drush_log(dt("PDF derivative creation succeeded for @pid.", array('@pid' => $object->id)), 'success');
+    }
+    else {
+      drush_log(dt("PDF derivative creation failed for @pid. Check the Drupal watchdog for detailed errors.", array('@pid' => $object->id)), 'error');
+    }
+  }
+}
```
adam-vessey, line 120: Should probably be setup as a batch...  Could possibly even `batch_set()` in an implmentation of `hook_update_N()` and have everything get migrated during the usual database update?

includes/admin.form.inc
```diff
@@ -105,3 +100,156 @@ function islandora_scholar_admin_form() {
...
+    }
+  }
+}
+
+/**
+ * Function updates PDF to text URL div.
+ */
+function islandora_update_pdftotext_url_div($form, $form_state) {
+  unset($form_state['submit_handlers']);
+  $form_state['rebuild'] = TRUE;
```
adam-vessey, line 252: I know similar code exists elsewhere... could probably still use a comment, though. `unset()`ing the submit handlers shouldn't be necessary? Maybe? Shrug?

includes/admin.form.inc
```diff
@@ -105,3 +100,156 @@ function islandora_scholar_admin_form() {
...
+  // AJAX wrapper for url checking.
+  $form['islandora_scholar_url_fieldset'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('TEXT'),
+  );
+
+  $form['islandora_scholar_url_fieldset']['islandora_scholar_create_fulltext'] = array(
+    '#type' => 'checkbox',
+    '#title' => t("Extract text streams from PDFs using pdftotext"),
+    '#description' => t("Extracted text streams are appended to PDFs as FULL_TEXT datastreams and are indexed into Solr. Uploading a text file takes priority over text stream extraction.
```
adam-vessey: Wording might be a bit off... Is text added to the PDF datastream, or a datastream on the object?

includes/admin.form.inc
```diff
@@ -105,3 +100,154 @@ function islandora_scholar_admin_form() {
...
+ *
+ * @param array $form
+ *   An array representing a Drupal form.
+ * @param array $form_state
+ *   An array representing the Drupal form state.
+ *
+ * @return array
+ *   An array representing a Drupal form.
+ */
+function islandora_scholar_admin_pdf_form($form, $form_state) {
```
adam-vessey: Should `form_load_include()` this file, because of the use of AJAX... Could end up loading the form from cache, and therefore not loading this file to have the AJAX callback available... Will have to change `$form_state` to be a reference.

PR #74, opened 2013-11-26
Stop leaking files on derivative creation.

jordandukart:


adam-vessey:
~~Merged this a bit early... Looks like we still leak the `fulltext.txt` files.~~ ~~Actually, I think it might be because I was poking at something else, which caused it to leave the file intact.~~

No, it was trying to run an ingest without having updated Islandora first...  And then it failed on trying to use a new core function. Great fun!

PR #76, opened 2013-12-06
Reference things correctly for export.

jordandukart:
Depends on https://github.com/Islandora/islandora_bookmark/pull/38

### DIFF
--- a/modules/bibliography/islandora_bibliography.module
+++ b/modules/bibliography/islandora_bibliography.module
@@ -215,8 +215,8 @@ function islandora_bibliography_preview_selected(array $form, array &$form_state
     }
   }
   if (count($pids)) {
-    $format = $form_state['input']['islandora_bookmark_export_options_select'];
-    $_SESSION['islandora_bibliography_preview']['style'] = $form_state['input']['islandora_bookmark_export_styles_' . $format];
+    $format = $form_state['values']['bookmarks']['fieldset']['export_options'];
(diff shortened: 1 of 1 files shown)


modules/bibliography/islandora_bibliography.module
```diff
@@ -215,8 +215,8 @@ function islandora_bibliography_preview_selected(array $form, array &$form_state
     }
   }
   if (count($pids)) {
-    $format = $form_state['input']['islandora_bookmark_export_options_select'];
-    $_SESSION['islandora_bibliography_preview']['style'] = $form_state['input']['islandora_bookmark_export_styles_' . $format];
+    $format = $form_state['values']['bookmarks']['fieldset']['export_options'];
+    $_SESSION['islandora_bibliography_preview']['style'] = $form_state['values']['bookmarks']['fieldset']['format_opts']['export_styles'][$format];
```
yqjiang, line 222: got Notice: Undefined index: default in islandora_bibliography_preview_selected() (line 219 of /var/www/drupal7/sites/all/modules/islandora_scholar/modules/bibliography/islandora_bibliography.module).

I dsm form state, no default in that array. may be should set $_SESSION['islandora_bibliography_preview']['style'] to a format when $format is defualt?

yqjiang, line 222: or just validate the form. force users to select format..
also may need also fix line https://github.com/jordandukart/islandora_scholar/blob/7.x/modules/bibliography/islandora_bibliography.module#L193

PR #89, opened 2014-02-26
csl 1.0.1 moved rnc to rng via Trang

willtp87:


modules/csl/data/apa.csl
```diff
@@ -1,367 +1,466 @@
...
+    <updated>2013-09-21T18:17:09+00:00</updated>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
   </info>
   <locale xml:lang="en">
-	<terms>
-	  <term name="translator" form="short">
-		<single>trans.</single>
-		<multiple>trans.</multiple>
-	  </term>
-	  <term name="ellipsis">...</term>
```
nigelgbanks, line 38: Hmm I'm slightly worried about the missing ellipsis, I know we added that for COH as they found bugs in the actual CSL style. In particular it was for citations with more than 7 authors.
http://blog.apastyle.org/apastyle/2009/08/formatting-apa-references-with-more-than-seven-authors.html

nigelgbanks, line 38: Seems as though the bug may have been fixed in 1.0.1
http://citationstyles.org/downloads/release-notes-csl101.html#id34

nigelgbanks, line 38: Works as correctly, even better than it used to actually, now it actually uses the ellipse character.

PR #103, opened 2014-04-11
Updated CiteProc php file to latest.

MorganDawe:


nigelgbanks:
We shouldn't be pulling in the .hg files from CiteProc into our repo, instead we should attempt to get our changes in our fork of CiteProc back into the main CiteProc, then implement CiteProc as a library.

PR #113, opened 2014-06-03
7.x Scholar Tests

qadan:
These tests test scholar! And all its sub-modules!

They also demonstrate that the Scholar module is currently broken in a few different ways. We should make tickets to resolve the existing problems before merging this.

This depends on https://github.com/Islandora/islandora/pull/507/files.


qadan:
Nah, the tests flag issues with the module things that should probably be made into other tickets. I'll ... figure this out soon, I'm still in vacation mode

qadan:
I'm going to remove this pull request for now; it's gotten so old that by the time we fix scholar and make the tests pass, my copy of the module is going to have to be rebased due to the crazy number of changes to scholar since then.

PR #117, opened 2014-08-06
7.x.select csl

willtp87:


js/citation_select.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * @file
+ * Lets the select ajax in a new citation.
+ */
+
+(function ($) {
+  Drupal.behaviors.islandora_scholar_citation_ajax = {
+    attach: function(context, settings) {
+      $(".islandora-scholar-citation-select").once("islandora-scholar-changei-csl", function() {
```
adam-vessey: Not really an issue... just looks weird: `islandora-scholar-changei-csl`

js/citation_select.js
```diff
@@ -0,0 +1,29 @@
...
+        $(this).change(function() {
+          var last_value = this.value;
+          $.ajax({
+            url: Drupal.settings.basePath + "islandora/object/" +
+              settings.islandora_scholar.pid + "/islandora_scholar_citation/" +
+              this.value,
```
adam-vessey: If a style name happens to contain a `/`, then this'll break.

includes/csl_select.form.inc
```diff
@@ -0,0 +1,38 @@
...
+ * @param array $form_state
+ *   The Drupal form state.
+ * @param string $pid
+ *   The PID of the object the citation will be generated for.
+ *
+ * @return array
+ *   The Drupal form definition.
+ */
+function islandora_scholar_citation_select_form(array $form, array &$form_state, $pid) {
+  drupal_add_js(array('islandora_scholar' => array('pid' => $pid)), 'setting');
```
adam-vessey: Not overly concerned, but any reason not to have this in the `#attached` bit below?
Sommat like:

``` php
  '#attached' => array(
    'js' => array(
      drupal_get_path('module', 'islandora_scholar') . '/js/citation_select.js',
      array(
        'type' => 'setting',
        'data' => array(
          'islandora_scholar' => array(
            'pid' => $pid,
          ),
        ),
      ),
    ),
  ),
```

js/citation_select.js
```diff
@@ -0,0 +1,29 @@
...
+  Drupal.behaviors.islandora_scholar_citation_ajax = {
+    attach: function(context, settings) {
+      $(".islandora-scholar-citation-select").once("islandora-scholar-change-csl", function() {
+        $(this).change(function() {
+          var last_value = this.value;
+          $.ajax({
+            url: Drupal.settings.basePath + "islandora/object/" +
+              settings.islandora_scholar.pid + "/islandora_scholar_citation/",
+            cache: true,
+            data: {"citation_style": encodeURIComponent(this.value)},
```
adam-vessey: Encoding here will end up double-encoding the value... Resulting in having to manually decode [where it's used](https://github.com/Islandora/islandora_scholar/pull/117/files#diff-6073e6790900fcfb2a74df790dee76bcR21).

PR #118, opened 2014-08-11
Add alter to allow modules to adjust citation info to be output.

adam-vessey:


modules/citeproc/citeproc.api.php
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Allow other modules to alter variables being output into citations.
+ *
+ * @param array $output
+ *   The associative array of variables being passed to citeproc.
+ * @param SimpleXMLElement $mods
+ *   The MODS from which the original output was produced.
+ */
+function hook_convert_mods_to_citeproc_jsons_alter(array &$output, SimpleXMLElement $mods) {
+
```
willtp87, line 19: Implementation please.

PR #130, opened 2014-10-06
Allow islandora_solr_metadata display for citation, thesis objects

JacobSanford:
Implements islandora_solr_metadata display for citation, thesis objects.

Checks to see if
- _islandora_solr_metadata_ is installed
- A field configuration has been created for the current content model
  and if both are true, renders the display with the user-configured solr metadata fields. Otherwise, it falls back to the old COINS method defined by this module.

There is some copy/paste code here between thesis/citation that could be refactored/expressed into more efficient methods, but I felt it best to keep hooks and templates completely separate for unique content models.

### DIFF
--- a/theme/islandora-citation.tpl.php
+++ b/theme/islandora-citation.tpl.php
@@ -1,0 +1,19 @@
+<?php
+/**
+ * @file
+ * This is the template file for the object page for citation objects.
+ */
+?>
+<div class="islandora-object-image">
+  <?php if (isset($variables['preview']) && isset($variables['pdf'])): ?>
+    <a href="<?php print $variables['pdf']; ?>">
+      <img src="<?php print $variables['preview']; ?>"/>
+    </a>
+  <?php endif; ?>
+</div>
+
+<?php if (isset($variables['metadata'])): ?>
+  <div class="citation_metadata">
+    <?php print $variables['metadata']; ?>
+  </div>
+<?php endif; ?>
--- a/theme/theme.inc
+++ b/theme/theme.inc
@@ -1,0 +1,34 @@
+<?php
+
+/**
+ * @file
+ * Theme hooks.
+ */
+
+/**
+ * Implements hook_preprocess().
+ */
+function islandora_entities_preprocess_islandora_citation(array &$variables) {
+  module_load_include('inc', 'islandora', 'includes/metadata');
+  $object = $variables['object'];
+  $object_pid = $object->id;
+  if ($object['PREVIEW'] && $object['PDF']) {
+    $variables['preview'] = "/islandora/object/$object_pid/datastream/PREVIEW/view";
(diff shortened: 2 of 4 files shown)


theme/theme.inc
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Implements hook_preprocess().
+ */
+function islandora_entities_preprocess_islandora_citation(array &$variables) {
+  module_load_include('inc', 'islandora', 'includes/metadata');
+  $object = $variables['object'];
+  $object_pid = $object->id;
+  if ($object['PREVIEW'] && $object['PDF']) {
+    $variables['preview'] = "/islandora/object/$object_pid/datastream/PREVIEW/view";
```
nigelgbanks, line 16: To support multi-sites that use sitename.com/multisite style URLs, could use the `url()` function to transform the internal paths to one that can be used with all types of multisites.

https://api.drupal.org/api/drupal/includes%21common.inc/function/url/7

theme/theme.inc
```diff
@@ -0,0 +1,34 @@
...
+    $variables['pdf'] = "/islandora/object/$object_pid/datastream/PDF/view";
```
nigelgbanks, line 17: Diddo `url()` function

theme/theme.inc
```diff
@@ -0,0 +1,34 @@
...
+ * Implements hook_preprocess().
+ */
+function islandora_entities_preprocess_islandora_thesis(array &$variables) {
+  module_load_include('inc', 'islandora', 'includes/metadata');
+  $object = $variables['object'];
+  $object_pid = $object->id;
+  if ($object['PREVIEW'] && $object['PDF']) {
+    $variables['preview'] = "/islandora/object/$object_pid/datastream/PREVIEW/view";
+    $variables['pdf'] = "/islandora/object/$object_pid/datastream/PDF/view";
+  }
```
nigelgbanks, line 32: Diddo url() function above lines 30-31.

theme/islandora-citation.tpl.php
```diff
@@ -0,0 +1,19 @@
+<?php
+/**
+ * @file
+ * This is the template file for the object page for citation objects.
+ */
+?>
+<div class="islandora-object-image">
+  <?php if (isset($variables['preview']) && isset($variables['pdf'])): ?>
```
nigelgbanks, line 8: You can use the variables directly like `$preview` instead of `$variables['preview']` Drupal uses [extract](http://php.net/manual/en/function.extract.php) to make all the $varibles properties into local variables.

nigelgbanks:
Aside from the notes above I think its' good to go, I'll give it a smoke test once the changes have been made. Thanks!

nigelgbanks:
It looks like this functionality is already present in the islandora_scholar module, it would seem that you first need to enable a setting in the administration
screen shot 2014-12-18 at 9 05 16 pm
 section to enable it though.

PR #133, opened 2014-10-06
Increase quality of PDF previews by controlling density used to generate...

JacobSanford:
Increase preview quality by allowing user control of density used. Defaulting to 600 significantly improves the text rendering in PDF->JPG.

screen shot 2014-10-06 at 8 25 22 pm
screen shot 2014-10-06 at 8 25 08 pm

### DIFF
--- a/includes/admin.form.inc
+++ b/includes/admin.form.inc
@@ -251,7 +251,7 @@ function islandora_scholar_admin_pdf_form($form, &$form_state) {
     '#title' => t('Density used to generate preview'),
     '#description' => t('The imagemagick density value used to generate the preview. Generally, higher numbers are better but require more computation time.'),
     '#element_validate' => array('element_validate_number'),
-    '#default_value' => variable_get('islandora_scholar_preview_density', 600),
+    '#default_value' => variable_get('islandora_scholar_preview_density', 75),
     '#size' => 5,
   );
   return system_settings_form($form);
--- a/includes/admin.form.inc
+++ b/includes/admin.form.inc
@@ -246,6 +246,14 @@ function islandora_scholar_admin_pdf_form($form, &$form_state) {
     '#default_value' => variable_get('islandora_scholar_preview_height', 700),
     '#size' => 5,
   );
+  $form['islandora_scholar_preview_fieldset']['islandora_scholar_preview_density'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Density used to generate preview'),
+    '#description' => t('The imagemagick density value used to generate the preview. Generally, higher numbers are better but require more computation time.'),
+    '#element_validate' => array('element_validate_number'),
+    '#default_value' => variable_get('islandora_scholar_preview_density', 600),
+    '#size' => 5,
+  );
   return system_settings_form($form);
 }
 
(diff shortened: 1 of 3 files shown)


dltj:
Thanks for making this change, Jacob.  Could I also ask you to add a line to the uninstall hook in islandora_scholar.install to remove the new variable?  Thanks!

JacobSanford:
Thanks for the feedback. Sure thing :)

willtp87:
This should default to the behaviour it had before.

includes/derivatives.inc
```diff
@@ -269,7 +269,7 @@ function islandora_scholar_create_jpg_derivative($file_uri, $dsid, $width, $heig
   // To make use of ImageMagick 6's parenthetical command grouping we need to
   // make the $source image the first parameter and $dest the last.
   // See http://www.imagemagick.org/Usage/basics/#cmdline
-  $command = escapeshellarg($source) . ' ' . implode(' ', $args) . ' ' . escapeshellarg("jpg:$dest");
+  $command = '-density ' . variable_get('islandora_scholar_preview_density', 600) . ' ' . escapeshellarg($source) . ' ' . implode(' ', $args) . ' ' . escapeshellarg("jpg:$dest");
```
willtp87: Can this be built up like the other args?

JacobSanford: Unfortunately not in this case : order matters with the density parameter - it has to come before the input file, and the other arguments are exploded after the filename. It could be included in a similar manner by creating a new array of pre-filename parameters, but that may be cumbersome for a single option.

willtp87: OK, it should likely still use http://php.net/manual/en/function.escapeshellarg.php

JacobSanford:
@willtp87 thanks for the feedback!

includes/admin.form.inc
```diff
@@ -246,6 +246,14 @@ function islandora_scholar_admin_pdf_form($form, &$form_state) {
     '#default_value' => variable_get('islandora_scholar_preview_height', 700),
     '#size' => 5,
   );
+  $form['islandora_scholar_preview_fieldset']['islandora_scholar_preview_density'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Density used to generate preview'),
+    '#description' => t('The imagemagick density value used to generate the preview. Generally, higher numbers are better but require more computation time.'),
+    '#element_validate' => array('element_validate_number'),
+    '#default_value' => variable_get('islandora_scholar_preview_density', 75),
```
willtp87: Isn't the default behaviour to use whatever is specified in the file with a fallback to 72? http://www.imagemagick.org/script/command-line-options.php#density I may very well be reading this wrong.

JacobSanford: Since PDF->image is usually vector rasterization, the density value is the DPI used by the Ghostscript delegate to rasterize the PDF. AFAIK vector PDFs have no density value, although the images contained within them may.

Honestly, I've always made the assumption that IM renders at 72.0 by default but never really looked into it. Good time, to do so, I guess! From the PDF delegate it references a core default resolution:

https://github.com/trevor/ImageMagick/blob/master/branches/ImageMagick-7.0.0/coders/pdf.c#L404-L405

Which does appear to be set to 72.0

https://github.com/trevor/ImageMagick/blob/master/branches/ImageMagick-7.0.0/MagickCore/image.c#L122

So, I do think that 72 is the default.

JacobSanford: I just noticed I used 75 instead of 72. My bad!

PR #139, opened 2014-10-16
Remove unused class and references to it.

jordandukart:


jordandukart:
Shouldn't require a release pull.

adam-vessey:
Just a note: The issue this code had been addressing has been fixed closer in to importers: https://github.com/Islandora/islandora_importer/blob/7.x/includes/importer.inc#L143-L174

PR #141, opened 2014-10-21
To fix the scope of the  variable defined in the lib/tcpdf/lang/*.php ...

matthewperry:
...files the file needs to be included not using module_load_include so that the language array defined stays in accessible scope.

### DIFF
--- a/modules/exporter/includes/pdf.inc
+++ b/modules/exporter/includes/pdf.inc
@@ -29,8 +29,11 @@
  *   The location to save the generated PDF.
  */
 function citation_exporter_create_pdf($html_input, $output_filename) {
-  require_once DRUPAL_ROOT . '/' . drupal_get_path('module', 'citation_exporter') . '/lib/tcpdf/config/lang/eng.php';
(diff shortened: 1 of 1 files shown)


modules/exporter/includes/pdf.inc
```diff
@@ -29,7 +29,7 @@
  *   The location to save the generated PDF.
  */
 function citation_exporter_create_pdf($html_input, $output_filename) {
-  module_load_include('php', 'citation_exporter', '/lib/tcpdf/config/lang/eng');
+  require_once DRUPAL_ROOT . '/' . drupal_get_path('module', 'citation_exporter') . '/lib/tcpdf/config/lang/eng.php';
```
adam-vessey: Not sure that `require_once()` is appropriate here... What would happen if `citation_exporter_create_pdf()` was called two (or more) times during the same request? Would `$l` be available to later invocations?

adam-vessey: Kinda thinking we could just add a `global $l;` line somewhere to import the global variable into our scope.

... Probably with a nice comment. :P

PR #164, opened 2015-03-18
Avoid building MODS edit form with islandora_scholar_add_usage_and_version_elements_to_mods() elements

JacobSanford:
The upload form (and methods) provided by includes/upload.tab.inc#[islandora_scholar_add_usage_and_version_elements_to_mods()](https://github.com/unb-libraries/islandora_scholar/blob/7.x/includes/upload.tab.inc#L212) add elements to the MODS datastream :

`<mods:accessCondition type="use and reproduction">author</mods:accessCondition>
<mods:physicalDescription authority="local">PUBLISHED</mods:physicalDescription>`

This form will attempt to render those elements when building the form, duplicating the PhysicalDescription block for input.

Adding XPath selectors to the avoids the READ method grabbing the unrelated elements on form build.

### DIFF
--- a/xml/thesis_form.xml
+++ b/xml/thesis_form.xml
@@ -787,7 +787,7 @@
               <value>physicalDescription</value>
             </create>
             <read>
-              <path>mods:physicalDescription</path>
+              <path>mods:physicalDescription[not(@authority) or @authority != 'local']</path>
               <context>parent</context>
             </read>
             <update>NULL</update>


JacobSanford:
Sure thing.

willtp87:
Spoke in commiters this may be rather backwards breaking and would need an update implementation if it were to go into the module.

JacobSanford:
I understand the concern.

I believe it is worth asserting strongly that this is a significant problem and needs be corrected ASAP. Not only are two identical fields (with different values being read/written) on a form confusing to a user, but (IMHO more importantly) any metadata submitted via that second form fieldset is deleted silently the next time _islandora_scholar_add_usage_and_version_elements_to_mods()_ is run:

https://github.com/unb-libraries/islandora_scholar/blob/7.x/includes/upload.tab.inc#L229

since that function deletes all children of [PhysicalDescription authority='local']. 

I'm admittedly unfamiliar with the ecosystem, but I'm not sure an update is possible here - since metadata is almost certainly compromised in most scholar installs because of this. To correct the broken MODS, an automated update hook would have to:
- Parse all Thesis MODS
- Extract the authority attributed elements and migrate them to the proper location
- Possibly compare or over-write elements provided by the 'first' form element that already exist. Adding them as multiples would not be effective, since the current form is designed to read a single element.

An effort could possibly be made to save the data from _islandora_scholar_add_usage_and_version_elements_to_mods()_ by moving elements from the authority='local' element to a local one. 

Anyhow, IMHO removing the form input doesn't create any problems that don't already exist (since not displaying the form element won't explicitly delete the MODS content, only avoid reading it), and the longer it is left in the more data insecurity is created for current installs going forward.

I'm willing to pitch in if you need additional help.

Regards,

JS

manez:
I am m.manez on Skype. Add me as contact and I'll add you to the Committer Call list :)

whikloj:
Ok, I played around with this a bit today and I am wondering if a good compromise can be achieved quickly. Jacob's point about people continuing to possibly add data to the wrong physicalDescription is valid. 

Perhaps if we add a second field to the form which selects any mods:physicalDescription that has an authority of local and has either a mods:form or mods:extent child element. 

If it could display them but not allow editing, perhaps with a warning that the content should be moved to the correct form element?

xpath="//mods:physicalDescription[@authority="local" and (string-length(mods:form/text()) &gt; 0 or string-length(mods:extent/text()) &gt; 0)]"

This in conjunction with Jacob's change should stop anyone from making the mistake while still letting people see (if not edit) their old content.

I was able to create a field to display the data in the form builder, but I couldn't figure out how to not display it for a new object. Perhaps someone with more XML form builder experience could help?

But I could be completely wrong too. :anguished:

mjordan:
7.x-1.5 has already been released, so closing.

PR #166, opened 2015-03-18
Ignore all of the Javascript for coder-review.

ruebot:


willtp87:
Was covered in commiters meeting, we should follow the same pattern as the other repos and only pull out the library code.

PR #169, opened 2015-03-19
Correct error in marcform authority element of thesis form

JacobSanford:
The physicalDescription element of the thesis MODS form contains an error that inserts incorrect metadata into MODS.


JacobSanford:
NP, glad to see them in HEAD.

PR #172, opened 2015-03-19
Adding a configurable max object export for pdf citations.

matthewperry:
Currently the pdf export cannot export a large PDF using TCPDF due it it running out of php max_execution_time during the execution of PDF generation.  For a temporary solution use a setting of 1200 objects that can be exported within the default 30 second execution time, this setting can be overridden manually using the citation export admin form so that if a server has a greater max execution time the object limit can be increased.


modules/exporter/citation_exporter.install
```diff
@@ -6,6 +6,16 @@
  */
 
 /**
+ * Implements hook_install().
```
willtp87: Wrong hook.

modules/exporter/citation_exporter.module
```diff
@@ -15,12 +15,22 @@ require_once dirname(__FILE__) . '/includes/batch.inc';
  */
 define('CITATION_EXPORTER_INDEX_SESSION', 'citation_exporter');
 define('CITATION_EXPORTER_EXPORT_MENU', 'islandora/citation/export');
+define('CITATION_EXPORTER_ADMIN_MENU', 'admin/islandora/tools/citation/export/config');
```
willtp87: I don't think this path needs to be so deep. Why not something more like `admin/islandora/tools/citaiton_export`

modules/exporter/citation_exporter.module
```diff
@@ -15,12 +15,22 @@ require_once dirname(__FILE__) . '/includes/batch.inc';
...
  */
 function citation_exporter_menu() {
   $items = array();
+  $items[CITATION_EXPORTER_ADMIN_MENU] = array(
+    'title' => 'Citation Exporter',
+    'description' => 'Configure citation exporter settings.',
+    'type' => MENU_NORMAL_ITEM,
+    'page callback' => 'drupal_get_form',
+    'page arguments' => array('citation_exporter_admin_form'),
+    'access arguments' => array('access administration pages'),
```
willtp87: `administer site configuration`?

modules/exporter/includes/admin.form.inc
```diff
@@ -0,0 +1,29 @@
...
+ */
+function citation_exporter_admin_form() {
+  $max_execution_time = (ini_get('max_execution_time') ? ini_get('max_execution_time') : "Unable to get max execution time.");
+  $form['citation_export_max_objects_for_pdf_export'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Maximium number of objects allowed per PDF citation export.'),
+    '#description' => t('The number of objects that can be exported to PDF,
+                        this number should be something that is achievable within the server\'s
+                        max execution time which is currently set to @execution_limit. For
+                        reference 30 seconds can handle up to 1200 objects.',
```
willtp87: This number is misleading without context.

modules/exporter/includes/admin.form.inc
```diff
@@ -0,0 +1,29 @@
...
+      array('execution_time' => $max_execution_time)),
+    '#element_validate' => array('element_validate_integer_positive'),
+    '#default_value' => variable_get('citation_export_max_objects_for_pdf_export', 1200),
```
willtp87: Var is not module prefixed.

modules/exporter/includes/batch.export.pdf.inc
```diff
@@ -25,7 +25,14 @@ function citation_exporter_batch_get_pdf_operations(array $pids, $style) {
   $set_id = citation_exporter_create_set($pids);
   $total = count($pids);
   $operations = array();
-
+  // XXX: Current limitations with TCPDF and time limits prevent the creation of
+  // a large PDF. For now to get around this the system will check the count of
+  // objects being exported and allow continue if it's less than the max number
+  // of pids that maintain a working PDF export.
+  if ($total > variable_get('citation_export_max_objects_for_pdf_export', 1200)) {
+    drupal_set_message(t('The number of selected objects exceeds the allowable PDF export limit.'), 'error', FALSE);
```
willtp87: This is probably ok, but maybe we should consider altering the form.

jordandukart:
I feel like all this code should be moved into [Islandora Bibliography](https://github.com/Islandora/islandora_scholar/blob/7.x/modules/bibliography/islandora_bibliography.module) as it is the one that implements the hooks for bookmark's export.

PR #177, opened 2015-03-26
loop hole in scholar embargo where POLICY DS can be deleted, meaning the...

nhart:
... object is not longer embargoed


nhart:
https://jira.duraspace.org/browse/ISLANDORA-1231

PR #179, opened 2015-04-01
Kylea/7.x

knjackson30974:


jordandukart:
Anything within libraries (CiteProc/tcpdf), .html or .xsl files shouldn't be modified as we don't maintain them explicitly.

PR #181, opened 2015-04-10
Ability to embargo multiple DSes for the same time period.

jordandukart:
Feature not a bug.


modules/islandora_scholar_embargo/includes/embargo.inc
```diff
@@ -294,15 +317,15 @@ function islandora_scholar_embargo_form_submit(array $form, array &$form_state)
...
-      if ($type === 'Object') {
-        $dsid = NULL;
-      }
-      else {
-        $dsid = $type;
+        if ($embargo_vals['type'] == 'Object') {
+          $dsid = NULL;
+        }
+        else {
+          $dsid[] = $type;
```
willtp87: Is now a misleading var name.

PR #184, opened 2015-04-14
Limit sending multiple e-mails about datastream embargos.

jordandukart:


modules/islandora_scholar_embargo/islandora_scholar_embargo.rules_defaults.inc
```diff
@@ -8,19 +8,29 @@
  * Implements hook_default_rules_configuration().
  */
 function islandora_scholar_embargo_default_rules_configuration() {
-  $datastream_expired_rule = rules_reaction_rule();
+  // @codingStandardsIgnoreStart
```
adam-vessey: These do not seem to be getting picked up, so [Travis is complaining](https://travis-ci.org/Islandora/islandora_scholar/jobs/58491354#L859)... Maybe just drop the exported rule into a file, `and file_get_contents()` it?

PR #198, opened 2015-07-16
Scholar creates physicalDescription fields that are overwritten by the default MODS form

whikloj:
Addresses [ISLANDORA-1350](https://jira.duraspace.org/browse/ISLANDORA-1350)


kimpham54:
just trying to recreate the issue right now with the original islandora_scholar module. if i enter in data in the physical description field, what error should I expect?  @whikloj

kimpham54:
@kimpham54

whikloj:
@kimpham54, as I recall what happens is if you add a PDF to the thesis it adds &lt;mods:PhysicalDescription @type="local"&gt; fields to the MODS record. However if you edit the MODS record, these fields are then returned and can be edited in the XML Form Builder MODS form. 
Then if you move from a pre-print to published copy of the thesis, those fields you may have added data to in the Form Builder are deleted and you will lose that data.

This PR "should":
1. No longer display the @type="local" physicalDescription fields when editing MODS.
and
2. Provide a separate disabled field to display this data so you can retrieve it and put it in the correct mods:physicalDescription fields.

kimpham54:
@whikloj 

This PR "should":
1. No longer display the @type="local" physicalDescription fields when editing MODS.
- tested, works. what looks to be the original physicalDescription field has the original text, the @type="local" field does not display when MODS record is first being entered
1. Provide a separate disabled field to display this data so you can retrieve it and put it in the correct mods:physicalDescription fields.
2. tested, if a new record is created with fix, disabled field is empty. if I had already created a record with the original islandora_scholar, it does display the information that was edited to the @type="local" for retrieval but it won't fix the poorly formed XML:

```
<mods:accessCondition type="use and reproduction">publisher</mods:accessCondition>
<mods:physicalDescription authority="local">PUBLISHED
<mods:form authority="marcform"/>
<mods:extent>edit extent second physicaldescription</mods:extent>
</mods:physicalDescription></mods:mods>
```

(physicalDescription cannot contain text)

kimpham54:
the pull request does do what you propose it should do though

whikloj:
Thanks @kimpham54!!

I'll see if I can reformat the previous XML to text for a better display. 

We don't store it (it is stripped out when you save) but it is better to have a nice display.

whikloj:
Actually, because the reason for that form field was in the case of people that had put text into those local physicalDescription fields with the old default MODS form.

Because you generated the MODS in the correct way, (ie. citation from scholar and regular from MODS form), you will not need access to the field. Unfortunately, there is no real way to tell if the contents of the field should be displayed.

I'll call upon the wisdom of @willtp87, as he knows alot about backwards compatibility stuff.

This field in the form builder is going to be useful only for people who have already messed up their data due to it getting stuck in the wrong fields. How important is the look of the form for people that are coming in with out concern for the old ways?

willtp87:
I'm not sure you'll want to wait on me, I probably won't be able to look at this until Tuesday or Wednesday.

willtp87:
I've been reading through this and if I understand this I see two options:
- keep things as they are in the pull regardless of the UX considerations as it keeps backwards compatibility which is the most important thing.
- Leave it up to the site admin by using a warning update hook to explain the situation prompting them to configure the use of a separate form that contains the `physicalDescription_bad`.  The form would be included but not associated, and we could remove it in the following release.

qadan:
FRESHLY MINTED COMPONENT MANAGER HEAD POKE IN okay go

We're kind of in a difficult spot here; we need those MODS statements for citeproc to function correctly, and likely to not break some sites' metadata displays. We likely have people with custom Scholar MODS forms that this wouldn't quite work for as well, making this extra difficult.

We can put in this fix for now, but we should put in a maintenance ticket for a long-term solution to this so we can discuss approaches to get rid of this problem altogether. Citeproc would have to be patched, which is scary but not impossible, but we'd also have to provide some weirdness to drive that part of the metadata display. Or something else. I dunno, this is super weird.

whikloj:
Thanks @willtp87 and @qadan, I like @willtp87's second option as it won't affect people unnecessarily but gives the solution if people have munged their metadata due to this problem. However I am also concerned that people might not be aware that their data is in the wrong physicalDescription fields.

So I'm going to leave this to @qadan. I am split on a resolution that is functional and not annoying.

whikloj:
Ok, perhaps I am overthinking this. I'll:
1. restore the original form except that it won't pick up physicalDescription fields with @authority='local'.
2. include the new form builder form (named thesis_form_FCREPO1350.xml).
3. include the clean up stylesheet (named cleanup_thesis_form_FCREPO1350.xslt).
4. add an update hook to warn about lost data in the &lt;mods:physicalDescription @authority='local' /&gt; element.

We probably need some documentation that can be pointed to from the update notice.

qadan:
I'm down to clown with due diligence messaging so long as we throw in a ticket to address this more thoroughly in the future. Which I did and it's here https://jira.duraspace.org/browse/ISLANDORA-1433

qadan:
HAHA WHOOPS

i wish there was a way in github to flag things as action items for a user instead of just 'mentioning' in a thread because sometimes i forget what's going on :disappointed:

PR #201, opened 2015-08-14
allow dynamic citations on thesis objects

bryjbrown:
Dynamic citations for thesis objects weren't working because 'islandora_scholar_citation_access' checks to make sure the object is an ir:citationCModel. Now it allows ir:thesisCModels as well. Tested on the 1.5 release VM.

### DIFF
--- a/islandora_scholar.module
+++ b/islandora_scholar.module
@@ -219,7 +219,7 @@ function islandora_scholar_romeo_access($object) {
  */
 function islandora_scholar_citation_access($object) {
   return variable_get('islandora_scholar_users_choose_display_csl', FALSE) &&
-    in_array(('ir:citationCModel' || 'ir:thesisCModel'), $object->models) &&
+    array_intersect(array('ir:citationCModel', 'ir:thesisCModel'), $object->models) &&
     isset($object['MODS']) &&
     islandora_datastream_access(ISLANDORA_VIEW_OBJECTS, $object['MODS']);
 }
--- a/islandora_scholar.module
+++ b/islandora_scholar.module
@@ -219,7 +219,7 @@ function islandora_scholar_romeo_access($object) {
  */
 function islandora_scholar_citation_access($object) {
   return variable_get('islandora_scholar_users_choose_display_csl', FALSE) &&
-    in_array('ir:citationCModel', $object->models) &&
+    in_array(('ir:citationCModel' || 'ir:thesisCModel'), $object->models) &&
     isset($object['MODS']) &&
     islandora_datastream_access(ISLANDORA_VIEW_OBJECTS, $object['MODS']);
 }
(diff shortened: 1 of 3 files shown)


bryjbrown:
@ruebot ticket created: https://jira.duraspace.org/browse/ISLANDORA-1389

islandora_scholar.module
```diff
@@ -219,7 +219,7 @@ function islandora_scholar_romeo_access($object) {
  */
 function islandora_scholar_citation_access($object) {
   return variable_get('islandora_scholar_users_choose_display_csl', FALSE) &&
-    in_array('ir:citationCModel', $object->models) &&
+    in_array(('ir:citationCModel' || 'ir:thesisCModel'), $object->models) &&
```
adam-vessey: Introducing an `OR` like this in PHP doesn't really work, I'm afraid. What I think you may want to do instead is an intersection, so something probably like:

``` php
array_intersect(array('ir:citationCModel', 'ir:thesisCModel'), $object->models)
```

If there's something in the intersection, the result will be a non-empty array, and a non-empty array has a truth value of `TRUE` in PHP.

I'm not sure if we want to address it now (or just make a ticket about it), but there seems to be other places where `ir:thesisCModel` wasn't introduced, such as the [RoMEO check](https://github.com/Islandora/islandora_scholar/blob/7.x/islandora_scholar.module#L209)... whether it should be or not... Shrug?

willtp87:
This is not strictly a bug. -- wrong

bryjbrown:
@ruebot Just let me know when you want the 7.x-1.6 and I'll take care of it on my end. Ditto for anything else related to this PR.

bryjbrown:
@willtp87 my rationale for filing the ticket as a bug was that if thesis objects weren't supposed to have the citeproc feature, then they shouldn't have the CSL selector dropdown available. The presence of the CSL dropdown implies that the citeproc stuff works for theses too.

willtp87:
Sorry, I see now. And the trouble's even with my own code. Maybe this pull should also remove this [if](https://github.com/Islandora/islandora_scholar/blob/7.x/includes/utilities.inc#L43)  in favor of an access check in [the form](https://github.com/Islandora/islandora_scholar/blob/7.x/includes/csl_select.form.inc#L23-L41) to keep these things from falling out of sync again.

bryjbrown:
Just made the changes @willtp87 suggested, seems to work on islandora_vagrant but I am a terrible, terrible Drupalist so I have no idea if this is the "right" way or not.

bryjbrown:
@ruebot Just to clarify, I should re-pull all new changes into my local repo, replicate the changes I made on 7.x to 7.x-1.6 and then re-push? And I should hold off on doing this until I get the :+1: from you?

qadan:
Sorry! I'm going to check on this today once I have a moment.

qadan:
Sorry again! This looks okay; merging

bryjbrown:
@ruebot Yep, I'm pretty jazzed about it! I'll make the release PR later today.

daniel-dgi:
@bryjbrown++

PR #202, opened 2015-08-17
Route Scholar Module Object Display Through Drupal's Theme Layer

Favenzio:
According to the instructions for Solution Pack development found here https://github.com/Islandora/islandora/wiki/Programming-Solution-Packs#display whenever hook_CMODEL_PID_islandora_view_object is implemented a call should be made to hook_theme in order to register the Solution Pack's output with the theme layer. This improvement grants Drupal themes the ability to override Islandora Scholar output using islandora-scholar-citation.tpl.php and islandora-scholar-thesis.tpl.php. See jira ticket https://jira.duraspace.org/browse/ISLANDORA-1391.


Favenzio:
@ruebot My mistake. I didn't want it closed. It should all be ok now.

Favenzio:
@ruebot No kidding. Thank you for helping me through it.  Owe you one. :beer:

jordandukart:
This is going to break any modules that implement `hook_islandora_object_view_alter()` that expect to interact with the previously existing renderable array. Similarly, the documentation linked in the corresponding JIRA ticket is incredible old and outdated. Finally, the other occurrences in core code where templates are being casted to an array is such that the underlying view_alters have a renderable array to work with, not the other way around.

Favenzio:
@jordandukart Thank you for responding to my pull request. Please excuse my ignorance regarding the staleness of the Islandora documentation I referenced. The page states that @qadan updated on May of 2014, which I didn't understand to be ancient. Are you aware of more recent documentation? I tried searching for some but was unable to find anything useful. Your closing comment mentions that there is core code currently doing what I suggest, namely, to have Islandora Scholar use Drupal's theme layer. Is there some advantage to circumventing it? I have worked with several solution packs (PDF, Large Image, Book) and they were all very easy to override thanks to their use of hook_theme. Those solution packs, however, do not implement their own get_view function. If you could point me to an example of the correct way of doing it I would be forever in your debt.

willtp87:
@Favenzio I have updated the information in the documentation.  Perhaps you could review it both for clarity and it may set you in the right direction.

Favenzio:
@willtp87 Thank you so much for updating the documentation so quickly. I'm trying to wrap my head around your changes, but I'm not sure if I understand correctly. You removed the line that states that a theme function 'must' be registered and instead noted that it is one possible approach. The rest of your change continues to explain how this would be done, and the example is the same one that was in the documentation before. Since that is the example I followed, it doesn't help me much. At the end you mentioned that the preferred way to write solution packs is to work with renderable arrays, and I don't have any problem with that. My problem is that the renderable array created in scholar is never registered with the theme layer, so there is no way to override its output in the theme. As far as I understand that is the preferred way for the presentation layer to be handled in Drupal. Am I incorrect about that?

Favenzio:
@ruebot Thank you for the invitation. I would be happy to attend if it would be a useful conversation to have, but I would hate to waste everyone's time if this is simply a lack of knowledge on my part. I'm doing a lot of research on the subject and I have found a couple of guides that are shedding some light on the problem I'm trying to tackle. Could you give me until tomorrow afternoon to research a solution? If I don't find something by then I think talking it out would be a great option.

willtp87:
@Favenzio I believe you can invoke the [view_object_alter](https://github.com/Islandora/islandora/blob/89507c5ce36a0766928bdf2b4df9f477fa7ebea5/islandora.api.php#L64-L88) hook in your template.php.

manez:
@Favenzio please do join us at the Committers Call! This week is specifically a call to discuss new features and improvements, so it's the perfect time to talk about this. Add me on Skype m.manez and I'l bring you in.

whikloj:
@Favenzio, you should definitely attend the committer's call. But I did exactly the same thing you did in trying to make the [compound solution pack block themeable](https://github.com/Islandora/islandora_solution_pack_compound/pull/83).

@willtp87 has got you on the right path, you can totally use that call in your template. I used a similar one for altering block content [in our template](https://github.com/uml-digitalinitiatives/UofM-Theme_2/blob/master/template.php#L489-L538).

Favenzio:
@willtp87 Thank you for pointing me in the direction of the view_object_alter hook. As @whikloj pointed out, that is probably what I need. I guess that's what @jordandukart what's getting at in his closing comment, I just completely missed it. Thanks for all your help guys.

Favenzio:
@ruebot I have added @manez to my Skype contacts. I will attend the committers meeting, although I no longer think this patch is necessary. Based on the information I read yesterday and today it seems Scholar is implementing its display in order to provide other Drupal module developers with the most flexibility, which is the Drupal best practice. Unfortunately, this makes theming more difficult (IMHO) but that seems to be a conscious decision made by the Drupal community a few years back.

PR #218, opened 2016-01-05
ISLANDORA-1552 - added ability for admins to specify use permissions and document versions available during PDF upload

joshdentremont:
In the admin settings you can now toggle whether to use the old values (used by default for backwards compatibility) or to specify your own. If selected you are able to add as many use permissions and document versions as you like, by entering them one per line in a text box. 

Whatever you enter will be both what is displayed to the user and what is stored in the metadata.

### DIFF
--- a/includes/pdf_upload.form.inc
+++ b/includes/pdf_upload.form.inc
@@ -20,30 +20,10 @@ function islandora_scholar_pdf_upload_form(array $form, array &$form_state) {
   $upload_size = min((int) ini_get('post_max_size'), (int) ini_get('upload_max_filesize'));
 
   // Pull document versions from the module settings if available.
-  if (variable_get('islandora_scholar_specify_document_versions', FALSE)) {
-    $document_version_names = array_map('trim', explode("\n", variable_get('islandora_scholar_document_versions', "Author's Original\nSubmitted Manuscript Under Review\nAccepted Manuscript\nProof\nVersion of Record\nCorrected Version of Record\nEnhanced Version of Record")));
-    $document_versions = array_combine($document_version_names, $document_version_names);
-  }
-  else {
-    $document_versions = array(
-      'PRE-PUBLICATION' => t('Pre-Publication'),
-      'PUBLISHED' => t('Published'),
-      'POST-PUBLICATION' => t('Post-Publication'),
-      'OTHER' => t('Other'),
-    );
-  }
+  $document_versions = get_document_versions();
 
   // Pull use permissions from the module settings if available.
-  if (variable_get('islandora_scholar_specify_use_permissions', FALSE)) {
-    $use_permission_names = array_map('trim', explode("\n", variable_get('islandora_scholar_use_permissions', "Publisher\nAuthor")));
-    $use_permissions = array_combine($use_permission_names, $use_permission_names);
-  }
-  else {
-    $use_permissions = array(
-      'publisher' => t('Contact Publisher (I do not hold copyright).'),
-      'author' => t('Contact Author (I hold the copyright and wish to retain all rights).'),
-    );
-  }
+  $use_permissions = get_use_permissions();
 
   $form['upload_pdf_checkbox'] = array(
     '#type' => 'checkbox',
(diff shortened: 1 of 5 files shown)


joshdentremont:
Sorry, I made a typo in the branch name. The JIRA ticket is 1552 not 1553.

https://jira.duraspace.org/browse/ISLANDORA-1552

includes/pdf_upload.form.inc
```diff
@@ -18,6 +18,33 @@
  */
 function islandora_scholar_pdf_upload_form(array $form, array &$form_state) {
   $upload_size = min((int) ini_get('post_max_size'), (int) ini_get('upload_max_filesize'));
+
+  // Pull document versions from the module settings if available.
+  if (variable_get('islandora_scholar_specify_document_versions', FALSE)) {
+    $document_version_names = array_map('trim', explode("\n", variable_get('islandora_scholar_document_versions', "Author's Original\nSubmitted Manuscript Under Review\nAccepted Manuscript\nProof\nVersion of Record\nCorrected Version of Record\nEnhanced Version of Record")));
+    $document_versions = array_combine($document_version_names, $document_version_names);
```
adam-vessey: Could get into trouble with translations here... Drupal variables can be translated... But the way it is done could lead to different languages having different sets of terms... there's no strong way to correlate between them.

Same below, with the `$use_permissions`.

joshdentremont:
@adam-vessey I'm not entirely sure what you mean. Is the issue that Author's Original, Submitted Manuscript Under Review, etc. are not going to be translated?

adam-vessey:
@J77D Not so much "not going to be" as "unable to be", but yes.

joshdentremont:
@adam-vessey Hmm... My thoughts were that everyone would be using their own list, and the one I specified was just a default that @dmoses provided. Perhaps he could weigh in on this? I'm far from an expert on MODS, but as I understand it, we're using physicalDescription authority="local" (https://www.loc.gov/standards/mods/userguide/physicaldescription.html) which I believe means that it is a non-standardized list? I guess if we are using the default list we should probably not be using authority="local" but instead note that it is a NISO standard?

For use permissions we're using accessCondition type="use and reproduction" (https://www.loc.gov/standards/mods/userguide/accesscondition.html) which looks like it is meant to contain whatever we deem appropriate so different languages will most likely have different terms anyway.

That being said, I am all for making the defaults translatable, I'm just unsure how to do it. I'm thinking adding $document_version_names = array_map('t', $document_version_names); would do it, but it looks like it is a security issue to translate user entered text https://api.drupal.org/api/drupal/includes!bootstrap.inc/function/t/7#sec_translating_vars. 

Maybe I'm over thinking it and it's as simple as changing all the variable_get lines to variable_get('islandora_scholar_document_versions', t("Author's Original\nSubmitted Manuscript Under Review\nAccepted Manuscript\nProof\nVersion of Record\nCorrected Version of Record\nEnhanced Version of Record")), but I'm not sure how smart t() is and whether it would deal with the \n properly?

Thoughts?

adam-vessey:
@J77D  `t()` isn't _that_ magical: It is meant only to deal with constant strings (with replacements).

... The variable itself as a whole should be able to be translated with ["Variable translation" (part of the i18n module)](https://www.drupal.org/node/1113374)... I'm just wanting to avoid painting ourselves into a corner... If we were using something like Drupal taxonomies for building the lists of terms (instead of parsing the content of a text area), then each term could be translated individually (and possibly play around with the level of indirection required to should the term in the correct language for display)... Dunno... Could possibly be added as a third option... I may be trying to borrow too much trouble from the(/a hypothetical) future.

joshdentremont:
@ruebot Currently it does what I had intended, which is to let me change the selections when uploading a PDF without having to change the code. I'm not sure what should be done about translation, if anything. If anyone has any requests I could spend a bit more time on this, but otherwise I'm OK with it as is. 

I see that I'm failing the Travis checks because of phpcpd. I'm not sure what would cause this, so any insight would be appreciated.

qadan:
My feeling is that there isn't any reason in the future that this _couldn't_ move to taxonomies, especially with the existence of get_document_versions() and get_document_permissions() as helper functions handling that part of it. This is a good stopgap solution for now; translation can be handled in the future if necessary.

PR #219, opened 2016-01-25
7.x islandora 1575 - escaping html in DOI importer

joshdentremont:
**JIRA Ticket**: https://jira.duraspace.org/browse/ISLANDORA-1575

Updates the DOI Importer to use DOMDocument instead of SimpleXMLElement to avoid issues with having to escape HTML.
# What's new?
- I've rewritten it to use DOMDocument, and import nodes rather than converting to strings and then creating new XML with those strings.
- During the change I noticed that it was reading in the language from the journal metadata and not using it so I removed that. https://github.com/Islandora/islandora_scholar/blob/7.x/modules/doi/includes/utilities.inc#L76
- I also removed any empty case statements which were not doing anything. I left in the ones with todo tags.
- It was previously not writing the genre, probably due to the case statement never being triggered (https://github.com/Islandora/islandora_scholar/blob/7.x/modules/doi/includes/utilities.inc#L228-L237). I added the genre at the top and set it to always fill in 'article' as described in https://wiki.duraspace.org/display/ISLANDORA/Islandora+Scholar#IslandoraScholar-CitationCollection&CitationStyleManagement
- Translated titles are now working. They were noted in the comments as not working previously. https://github.com/Islandora/islandora_scholar/blob/7.x/modules/doi/includes/utilities.inc#L162-L170
# How should this be tested?

Compare the MODS generated for the same DOI before and after these changes.

I tested the following DOIs:
10.1111/asap.12106
10.1007/bf02328068
10.1021/acs.est.5b01589
10.1111/1745-9125.12034
10.3989/revmetalm.1998.v34.iextra.709
10.1021/acs.est.5b01589

Each one returned the same XML as they did with the old code (in a few places, the ordering of elements may have changed), except the new ones had the genre and did not break with special characters (ampersands, etc.). 
# Additional Notes:
- The function islandora_doi_transform_to_mods should take a DOMDocument but it will also work with a SimpleXMLElement.
# Interested parties

@Islandora/7-x-1-x-committers


qadan:
I'm definitely on board with fixing this issue! Just a touch concerned about the implementation; it looks like we'd be doing two steps of en/decoding to get to something that should take zero steps (i.e. decoding from a simplexml node during the string casting process and encoding back using htmlspecialchars()).

Honestly, we should be getting away from using PHP simplexml just due to the number of times we run into silly errors like this one, and how difficult they are to fix just using simplexml itself.

It would be my opinion that we should be loading the original CrossRef data as a DOMDocument object and avoid special character wrangling altogether

joshdentremont:
@ruebot Oops, I originally read that comment as a call for opinions. 

If I rewrite the islandora_doi_transform_to_mods function to take and return a DOMDocument instead of a SimpleXMLElement it looks like I can use DOMDocument's importNode function to copy the xml elements from one to another, avoiding those casts to strings. @qadan Please let me know if that's not what you meant.

I will try this out in the (hopefully) not too distant future.

qadan:
That's pretty much it, yep! Only thing we should note (not to pile too many things onto this) is that if we're changing the function signature, we should also put this through a deprecation cycle - namely accepting anything as the XML parameter, checking the type the parameter, and if it's simplexml, converting it to DOMDocument and popping a warning. I was looking at [islandora_deprecated()](https://github.com/Islandora/islandora/blob/14c8c68b4dcce4aca3f578e0daad4c2aa1951bed/includes/utilities.inc#L808) to accomplish this, but the wording it uses implies that the entire function is being deprecated; I don't think we have any helpers to pop a warning on function signature being changed.

joshdentremont:
I've played around with this and I don't think we're going to be able to avoid reading the text as a string and re-encoding it. 

Using DOMDocument::importNode allows us to copy an element and preserves encoding (for ampersands, etc.), but we can't change the type. For example, we are copying full_title from the CrossRef XML to our MODS XML, but in MODS it needs to be called title

If we create a new DOMNode and just copy the text from the node it converts &amp;amp; to & and we are stuck using something like htmlspecialchars() to fix it. I don't think this is any better than using SimpleXML as we are currently doing.

I can still rewrite this using DOMDocument instead of SimpleXML, but I don't think it will help anything.

As an example, have a look at the following code and its output (if you uncomment the line with htmlspecialchars it works as expected):

```
<?php
$old = new DOMDocument('1.0');
$new = new DOMDocument('1.0');

// Create original XML
$old->loadXML('<root><inner>Text &amp; Ampersand</inner></root>');
$inner = $old->getElementsByTagName("inner")->item(0);

// Clone a node from the original
$imported = $new->importNode($inner, true);
$new->appendChild($imported);

echo "Original:\n" . $old->saveXML() . "\n";

echo "Imported:\n" . $new->saveXML() . "\n";

// Copy the node contents but change the element type
$new = new DOMDocument('1.0');
$changed = $new->createElement('root', $inner->nodeValue);
// $changed = $new->createElement('root', htmlspecialchars($inner->nodeValue));
$new->appendChild($changed);

echo "Changed title:\n" . $new->saveXML();
?>

Original:
<?xml version="1.0"?>
<root><inner>Text &amp; Ampersand</inner></root>

Imported:
<?xml version="1.0"?>
<inner>Text &amp; Ampersand</inner>


Warning: DOMDocument::createElement(): unterminated entity reference       Ampersand in test.php on line 19
Changed title:
<?xml version="1.0"?>
<root/>
```

adam-vessey:
It appears that separately creating and appending the `DOMText` nodes does the trick.

I just ran this in the `devel/php` dealio:

``` php
$doc = new DOMDocument();
$element = $doc->createElement('asdf');
$text = $doc->createTextNode('this & that');
$element->appendChild($text);
$doc->appendChild($element);
dsm($doc->saveXML());
```

and it output as expected:

``` xml
<?xml version="1.0"?>
<asdf>this &amp; that</asdf>
```

Similarly, avoiding going out to the `nodeValue` can avoid the issue, by importing the `DOMText` nodes directly:

``` php
$doc = new DOMDocument();
$element = $doc->createElement('asdf');
$text = $doc->createTextNode('this & that');
$element->appendChild($text);
$doc->appendChild($element);
dsm($doc->saveXML());

$import = new DOMDocument();
$element = $import->createElement('import');
foreach ($doc->documentElement->childNodes as $child) {
  $imported = $import->importNode($child);
  $element->appendChild($imported);
}
$import->appendChild($element);
dsm($import->saveXML());
```

yielding:

``` xml
<?xml version="1.0"?>
<asdf>this &amp; that</asdf>
```

and

``` xml
<?xml version="1.0"?>
<import>this &amp; that</import>
```

adam-vessey:
Also, just noticed the [notes on `DOMDocument::createElement()`](http://php.net/manual/en/domdocument.createelement.php#refsect1-domdocument.createelement-notes): "The value will not be escaped. Use DOMDocument::createTextNode() to create a text node with escaping support. "

joshdentremont:
Thanks Adam. I'll give that a go.

joshdentremont:
I've rewritten it to use DOMDocument, and import nodes rather than converting to strings and then creating new XML with those strings. 

During the change I noticed that it was reading in the language from the journal metadata and not using it so I removed that. https://github.com/Islandora/islandora_scholar/blob/7.x/modules/doi/includes/utilities.inc#L76

I also removed any empty case statements which were not doing anything. I left in the ones with todo tags.

It was previously not writing the genre, probably due to the case statement never being triggered (https://github.com/Islandora/islandora_scholar/blob/7.x/modules/doi/includes/utilities.inc#L228-L237). I added the genre at the top and set it to always fill in 'article' as described in https://wiki.duraspace.org/display/ISLANDORA/Islandora+Scholar#IslandoraScholar-CitationCollection&CitationStyleManagement 

I tested the following DOIs:
10.1111/asap.12106
10.1007/bf02328068
10.1021/acs.est.5b01589
10.1111/1745-9125.12034 
10.3989/revmetalm.1998.v34.iextra.709 
10.1021/acs.est.5b01589

Each one returned the same XML as they did with the old code (in a few places, the ordering of elements may have changed), except the new ones had the genre and did not break with special characters (ampersands, etc.). Also, translated titles are now working. They were noted in the comments as not working previously. https://github.com/Islandora/islandora_scholar/blob/7.x/modules/doi/includes/utilities.inc#L162-L170

bryjbrown:
@J77D @dannylamb @DonRichards I don't have any experience with the DOI importer so I was waiting for someone else to test and merge this, but I want to make sure this PR makes it into 1.9 so I'm testing it myself.

I loaded up a fresh vagrant box and fetched this PR banch, but both the PR branch *and* 7.x seem to be having the same problem for me: when I'm in a collection and go to Manage -> Collection -> Batch Import Objects -> DOI Importer and upload either a file of DOIs *or* paste in some DOIs into the textfield and click "Import", it starts a batch process that only lasts for a fraction of a second and  then ends with a success message. However, no items are actually imported AFAICT. I'm using the DOIs @J77D provided above, so I can't tell if the issue is that I'm doing it wrong, or perhaps theres a networking issue with VM. Anybody have any thoughts on this so I can complete a quick test and finally get this PR merged?

joshdentremont:
@bryjbrown Did you update the settings at admin/islandora/solution_pack_config/scholar/islandora_doi? You need to sign up for a crossref account if you don't have one.

Thanks for testing this.

DonRichards:
@bryjbrown When I tested this a while back I got the same. I'll see if I can do this test again.

bryjbrown:
@J77D Thanks for the help, I now have DOIs importing correctly (I think @DonRichards and I fell victim to [ISLANDORA-1098](https://jira.duraspace.org/browse/ISLANDORA-1098)).

Its worth noting that the DOIs provided didn't have ampersands in the title so they weren't a real test of the PR, but I found some DOIs that do:
- 10.1107/S0907444998003254
- 10.1111/j.1532-5415.1991.tb01616.x

I include these in case @DonRichards wants to test this PR as well. They worked just fine for me, outputting the proper title in the PR when they had no title in the 7.x branch. I approve of this PR, and pending @DonRichards blessing I say we merge it. 👍 

Thanks for your work on this, @J77D and apologies for the long overdue review.

EDIT: Looks like there are merge conflicts to be resolved before we can merge. GitHub recommends doing the following:

# Step 1: From your project repository, check out a new branch and test the changes.
- git checkout -b J77D-7.x-ISLANDORA-1575 7.x
- git pull https://github.com/J77D/islandora_scholar.git 7.x-ISLANDORA-1575

# Step 2: Merge the changes and update on GitHub.
- git checkout 7.x
- git merge --no-ff J77D-7.x-ISLANDORA-1575
- git push origin 7.x

@dannylamb Can you advise on if this fits our standard workflow, or if we should do it a different way?

bryjbrown (approved):

DonRichards:
Looks good to me as well. Thanks for the effort @J77D and @bryjbrown

DonRichards:
@bryjbrown I would merge it but I don't have access to that for some reason.

DiegoPino:
@J77D @bryjbrown this pull has some conflicts and can't be merged "automagically". Since a year already passed, for sure new code got in the way and you could need a rebase or some cherry picking 😬 to get it into shape again. Please ping me if you need help on that. 😄  Thanks!

joshdentremont:
Looks like the conflict is the change here: https://github.com/Islandora/islandora_scholar/commit/5613c83849a5459084256ff1396acaf74802d6fd

I believe it is fine for my change to override Jonathan's change, since we are no longer using SimpleXMLElement and the line he changed has been removed.

DiegoPino:
@J77D yeah, the issue is not the change or the code itself you are pulling, but how git decides where to put your code, simply the diff is not matching, so you need to rebase or cherry pick.

joshdentremont:
Ok, It should be good to merge now.

Thanks @DiegoPino, @bryjbrown, and @DonRichards for your help

bryjbrown:
Travis is all green, so we're in business. PR 219 is officially merged. Thanks for taking the initiative to straighten out the merge conflict, @J77D. :1st_place_medal:

PR #220, opened 2016-02-04
Enables roles with "can embargo any object" to manage embargoed items

bryjbrown:
Currently embargoes are implemented by applying XACML policies to objects restricting access to all but administrators and the object's owner. The permission "can embargo any object" provided by the module has the description "User can add or remove embargo on any object in repository", but this is not the case since any role granted this permission is unable to access the object due to the user-centric XACML policy. This pull request adds a check for all roles with the "can embargo any object" permission and adds them to the XACML policy when setting embargoes.

Use case: FSU has several staff at the graduate school who need to be able to manage embargoes independently on ETDs. We created a low-power "embargo_admin" role with just these permissions, and this pull request was required in order for that role to be able to see embargoed objects.

### DIFF
--- a/modules/islandora_scholar_embargo/islandora_scholar_embargo.module
+++ b/modules/islandora_scholar_embargo/islandora_scholar_embargo.module
@@ -479,8 +479,8 @@ function islandora_scholar_embargo_set_embargo($param, $dsid = NULL, $end = 'ind
       foreach ($users as $user) {
         $xacml->viewingRule->addUser($user->name);
       }
-      foreach ($roles as $key => $name) {
-        $xacml->viewingRule->addRole($name);
+      foreach ($roles as $key => $role) {
+        $xacml->viewingRule->addRole($role);
       }
     }
   }
@@ -512,8 +512,8 @@ function islandora_scholar_embargo_set_embargo($param, $dsid = NULL, $end = 'ind
         foreach ($users as $user) {
           $xacml->datastreamRule->addUser($user->name);
         }
-        foreach ($roles as $key => $name) {
-          $xacml->datastreamRule->addRole($name);
+        foreach ($roles as $key => $role) {
+          $xacml->datastreamRule->addRole($role);
         }
       }
     }
--- a/modules/islandora_scholar_embargo/islandora_scholar_embargo.module
+++ b/modules/islandora_scholar_embargo/islandora_scholar_embargo.module
@@ -462,6 +462,9 @@ function islandora_scholar_embargo_set_embargo($param, $dsid = NULL, $end = 'ind
   $xacml_changed = FALSE;
 
   $users = islandora_scholar_embargo_users_to_notify($item);
+  $roles = user_roles(FALSE, 'can embargo any object');
+  // ISLANDORA_SCHOLAR_EMBARGO_CAN_EMBARGO_ANY
+
   if ($dsid === NULL) {
     $item->relationships->remove(ISLANDORA_SCHOLAR_EMBARGO_RELS_URI, ISLANDORA_SCHOLAR_EMBARGO_EXPIRY_PRED);
     $item->relationships->remove(ISLANDORA_SCHOLAR_EMBARGO_RELS_URI, ISLANDORA_SCHOLAR_EMBARGO_NOTIFICATION_PRED);
@@ -472,9 +475,12 @@ function islandora_scholar_embargo_set_embargo($param, $dsid = NULL, $end = 'ind
       if ($notification !== NULL) {
         $item->relationships->add(ISLANDORA_SCHOLAR_EMBARGO_RELS_URI, ISLANDORA_SCHOLAR_EMBARGO_NOTIFICATION_PRED, $notification, $type);
       }
+      $xacml_changed = TRUE;
       foreach ($users as $user) {
         $xacml->viewingRule->addUser($user->name);
-        $xacml_changed = TRUE;
+      }
+      foreach ($roles as $key => $name) {
+        $xacml->viewingRule->addRole($name);
       }
     }
   }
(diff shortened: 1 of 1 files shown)


manez:
Thanks for this, @bryjbrown . To get all of our ducks in a row, would you mind creating JIRA ticket to go with this pull? To merge, we'll also need either an [iCLA](http://islandora.ca/sites/default/files/islandora_cla.pdf) for you or a [cCLA](http://islandora.ca/sites/default/files/islandora_ccla.pdf) from FSU, depending on whether this is all your own work or done on behalf of FSU. Thanks!

bryjbrown:
@manez I created the ticket: https://jira.duraspace.org/browse/ISLANDORA-1602

You should have my contributor agreement on file, I did a pull request last year for Scholar and had to fill one out.

manez:
@bryjbrown Oops! You are entirely correct! You have a CLA on file. Carry on, nothing to see here...

(thanks for the JIRA ticket)

jordandukart:
Hey @bryjbrown,

This pull has ramifications that go beyond allowing users with roles to be able to access/manage embargoed items. The original design choice was to account for individual users only. Consider the use case where a new user role is created and given the `ISLANDORA_SCHOLAR_EMBARGO_CAN_EMBARGO_ANY` permission. At such point all existing embargoed objects' XACMLs have to be updated to account for the addition of this role. This would be very expensive to traverse a repository full of embargoes as such action takes place. Perhaps your use case could be satisfied by adding default configuration of specific users and roles to embargo similar to how the XACML module provides through its admin page? Any addition of roles to embargo will have to have additional code implemented to notify by role for embargo expiry/update e-mails.

## Thanks!

Jordan Dukart
_Developer_
**[discoverygarden inc.](http://www.discoverygarden.ca) | Managing Digital Content**

bryjbrown:
@jordandukart If the original design choice was to account for individual users only, then what purpose does it serve to assign the `ISLANDORA_SCHOLAR_EMBARGO_CAN_EMBARGO_ANY` permission to roles? 

I'm fully aware of how expensive it is to re-apply the XACML policy to all embargoed items, as this is exactly what I had to do when I found the error. 

If its an intentional design choice to not allow access by roles, then the permissions offered are very misleading.

rosiel:
Hi @bryjbrown, as discussed in the committers call we think that this needs a solution, but we came up with a slightly different approach to avoid the need for future modifications to XACML datastreams. 

https://jira.duraspace.org/browse/ISLANDORA-1602?focusedCommentId=48742

If we programmatically create a role and give it the permission `ISLANDORA_SCHOLAR_EMBARGO_CAN_EMBARGO_ANY`, then we can add people to this role (and/or any other role) without needing to update the XACML datastreams. Except the first time, of course, which would insert this role into all existing embargoed XACML datastreams. If this pull request also included a drush script to do that, it'd be golden.

bryjbrown:
@rosiel Okay, just repeating to make sure I understand:
1. We should throw out this current PR and start a new one since the committers want a different solution
2. New solution involves programmatically creating an `embargo_master` role and the XACML policy applied by embargoes should allow access to this role.
3. There should be a drush script that seeks out all existing embargoed objects and adds access for `embargo_master` to the policy.
4. Remove `ISLANDORA_SCHOLAR_CAN_EMBARGO_OWNED` and `ISLANDORA_SCHOLAR_CAN_EMBARGO_ANY` parts of the code since embargo management privs will be given by adding the `embargo_master` role to existing users.

Apologies if this is redundant, I just want to make sure I'm interpreting the requirements correctly.

rosiel:
1. You could edit the current branch, since you've already started it, or start a new PR as you wish. 
2. Yep - and you can name the role whatever you like, I called it that because i needed a placeholder to refer to it as. 
3. Yes. You may have had something similar if you implemented the solution in this current PR.
4.  On the Drupal side, we still need the `ISLANDORA_SCHOLAR_CAN_EMBARGO_ANY` permission so that users can access the embargoed object interfaces. This is the permission that should be granted to the `<scarequotes>`embargo_master`</scarequotes>` role. The permission `ISLANDORA_SCHOLAR_CAN_EMBARGO_OWNED` hasn't been mentioned yet in this issue, but it likely has the same issue in that changing who has this permission will not affect their ability to edit/see existing objects. If I understand it right, it's impossible to fix the this one without triggering XACML rewrites every time this permission changes. Tagging @Islandora/7-x-1-x-committers  for suggestions.

bryjbrown:
@rosiel One problem I forgot to mention earlier that would affect the way this problem gets solved is a potential security issue with having a single "embargo_master" role in a multisite context. At FLVC we have multiple institutions' objects in a single fedora separated by namespace, but due to some collection sharing we do allow one institution's Islandora multisite instance to access objects in other institutions' namespaces. For instance, Florida State's interface can access Florida Atlantic's objects and vice versa.

For this reason we typically namespace roles as well, such as fsu_collection_manager vs fau_collection_manager. We followed the same rule with the embargo admin role, creating an "fsu_embargo_admin" role so that only FSU folks can set embargoes on FSU objects. If an object has "fsu_embargo_admin" in the XACML policy, someone with the "fau_embargo_admin" can't mess with them. Contrast this with having one universal "embargo_admin" role accross institutions where a person at FAU with the "embargo_admin" role could edit FSU objects. This is why the current pull request is the way it is; it was deemed a more secure option by FLVC, the ones who maintain this multisite instance.

If the committers deem this an edge case not worth worrying about, then I can proceed with the single "embargo_admin" plan as specified earlier, but I do think this is something worth considering since multisite instances seem to be on the rise. Having the Drush script to reset all existing embargoes added to the current pull request another option that would result in a complete solution to the problem, too.

rosiel:
Thanks @bryjbrown! I think that's a pretty important use case, and is a serious flaw in the "solution" I proposed. I'm personally thinking your original solution might be the way to go... Can you make it to committers call on Thursday to discuss it?

bryjbrown:
@rosiel Yeah, I'll be at the committers call. Going to try to sit in on those regularly from now on when possible :+1:

bryjbrown:
Since this was superceded by #232 (merged), I'm closing this PR.

PR #230, opened 2016-04-08
strtolower roleTerm to address ISLANDORA-1657

nhart:
**JIRA Ticket**: https://jira.duraspace.org/browse/ISLANDORA-1657

It applies strtolower to the roleTerm
# How should this be tested?

Replace the MODS DS of a citation object with the supplied MODS XML (in the ticket)
Prior to the PR the citation will be: 
Last, F. (2014). Title about some citation. Some other title. Retrieved from http://someurl.com 

After applying the PR, the citation will be: 
Hart, N., & Last, F. (2014). Title about some citation. Some other title. Retrieved from http://someurl.com 
# Interested parties

tagging @Islandora/7-x-1-x-committers as maintainer is internal.

### DIFF
--- a/modules/citeproc/includes/converter.inc
+++ b/modules/citeproc/includes/converter.inc
@@ -553,6 +553,7 @@ function convert_mods_to_citeproc_json_names(SimpleXMLElement $mods) {
     $names = $mods->xpath($path);
     if (!empty($names)) {
       foreach ($names as $name) {
+
         add_mods_namespace($name);
         $role = convert_mods_to_citeproc_json_name_role($name, $valid_roles, $default_role);
         if ($role !== FALSE) {
@@ -677,7 +678,7 @@ function convert_mods_to_citeproc_json_name_role(SimpleXMLElement $name, array $
   module_load_include('inc', 'citeproc', 'includes/marcrelator_conversion');
   $role_term = $name->xpath("mods:role/mods:roleTerm");
   if (isset($role_term[0])) {
-    $role = (string) $role_term[0];
+    $role = strtolower((string) $role_term[0]);
     $role_term = $role_term[0];
   }
   else {


rosiel:
@nhart ++ This is something we ran into. I'll test.

rosiel:
It works. And it opens a whole can of worms if you actually use it!

The default citation form shipped with Scholar writes `<name type='personal'><role><roleTerm authority="marcrelator" type="text">author</roleTerm></role></name>` and reads `mods:name[@type = 'personal' and mods:role/mods:roleTerm = 'author']`. If you input a record with the correct marcrelator term 'Author', the name won't show up when you try to edit the citation. A solution could be to change the read xpath to `mods:name[@type = 'personal' and (mods:role/mods:roleTerm = 'author' or mods:role/mods:roleTerm = "Author")]` which will pick up both, but it will still update `self::node()` so you'll end up with potentially a mix of author and Author in your resulting metadata. I would suggest that we do this at least so that it's not quite so broken. It's worth a warning that if you're going to use this new feature and you've customized your form, you'll have to update your read query.

I was worried about transforming to solr, too. But - yay! slurp_all_MODS_to_solr.xslt lowercases the roleterm. https://github.com/discoverygarden/basic-solr-config/blob/modular/islandora_transforms/slurp_all_MODS_to_solr.xslt#L170

Update: pull request added. https://github.com/nhart/islandora_scholar/pull/1

nhart:
In theory, if you have roleTerm "Author" in your XML, your form (not the built in one) is already set up to handle this. This isn't changing the default behavior.

rosiel:
Fair enough. Though you could also get Author by importing correct MODS records. But if that was your intended use case, I'm happy to merge after the 24h have passed. 

[Edit: MODS not MADS]

PR #232, opened 2016-04-20
embargo roles whitelist

qadan:
**JIRA Ticket**: https://jira.duraspace.org/browse/ISLANDORA-1602
- Other Relevant Links: https://github.com/Islandora/islandora_scholar/pull/220 (the original pull request for this issue)

Allows a roles whitelist to be configured and applied to embargoes.
# What's new?

Three things:
- A configurable whitelist has been added to establish trusted roles with the 'can embargo any' permission that are added to embargoes when they are set via `islandora_scholar_embargo_set_embargo()`.
- A batch has been added to reapply existing embargoes. It can be triggered via the configurable whitelist form, if desired.
- A drush command has been added, `islandora_scholar_embargo_reapply_embargoes`, which also triggers the batch re-application.
# How should this be tested?
- The whitelist form should be tested to make sure the variables it sets take the form of an associative array pairing role IDs with role names.
- The batch should be tested to ensure embargoes on objects and datastreams are applied properly
- The drush command should also be tested to ensure it functions as expected.

As this has XACML security implications, we should be ensuring that when an embargo is created, a POLICY is applied that _only_ contains the roles in the whitelist. POLICYs for entire objects should have viewing _and_ management permissions for whitelisted roles.
# Additional Notes:

Care has been taken to structure the descriptions of things so that there's no ambiguity, since ambiguous and misleading documentation seems to be part of the source of this issue. Documentation within the module should probably be checked to make sure it's internally consistent and communicates clearly.

One of the things that is documented in this pull but bears repeating is that due to the way the Fedora-Drupal filter works, matching role IDs between sites that share the same Fedora represent a method of seeing objects that would otherwise be hidden. Whitelist roles should likely be unique and/or namespaced to mitigate this issue.
# Interested parties

From the original PR: @bryjbrown @rosiel @jordandukart

### DIFF
--- a/modules/islandora_scholar_embargo/includes/admin.form.inc
+++ b/modules/islandora_scholar_embargo/includes/admin.form.inc
@@ -21,7 +21,9 @@ function islandora_embargo_admin(array $form, array &$form_state) {
   $options = islandora_get_content_models();
   $selected = variable_get('islandora_embargo_content_models', array('ir:citationCModel', 'ir:thesisCModel'));
   foreach ($selected as $cmodel) {
-    $options = array($cmodel => $options[$cmodel]) + $options;
+    if (isset($options[$cmodel])) {
+      $options = array($cmodel => $options[$cmodel]) + $options;
+    }
   }
   foreach ($options as $key => $value) {
     $rows[$key] = array(
--- a/modules/islandora_scholar_embargo/includes/batch.inc
+++ b/modules/islandora_scholar_embargo/includes/batch.inc
@@ -156,7 +156,8 @@ function islandora_scholar_embargo_reapply_embargoes_batch_operation(&$context) 
   $sandbox =& $context['sandbox'];
   if (empty($sandbox)) {
     $sandbox['completed'] = 0;
-    $sandbox['offset'] = '1900-01-01T00:00:00.000Z';
+    $sandbox['date_offset'] = '1900-01-01T00:00:00.000Z';
+    $sandbox['pid_offset'] = NULL;
   }
 
   // It's possible for this number to change during the batch (e.g., if an
@@ -169,7 +170,7 @@ function islandora_scholar_embargo_reapply_embargoes_batch_operation(&$context) 
   }
   else {
     // Iterate through a set.
-    $set = islandora_scholar_embargo_get_embargoed_items_subset($sandbox['offset']);
+    $set = islandora_scholar_embargo_get_embargoed_items_subset($sandbox['date_offset'], $sandbox['pid_offset']);
     if (!empty($set)) {
       foreach ($set as $embargoed_item) {
         $id_parts = explode('/', $embargoed_item['obj']['value']);
@@ -179,7 +180,8 @@ function islandora_scholar_embargo_reapply_embargoes_batch_operation(&$context) 
         islandora_scholar_embargo_lift_embargo($pid, $dsids);
         islandora_scholar_embargo_set_embargo($pid, $dsids, $embargoed_item['expiry']['value']);
         $sandbox['completed']++;
-        $sandbox['offset'] = $embargoed_item['date']['value'];
+        $sandbox['date_offset'] = $embargoed_item['date']['value'];
+        $sandbox['pid_offset'] = $pid;
         $context['message'] = t('@count/@total: Re-applied embargo to @item of @pid.', array(
           '@count' => $sandbox['completed'],
           '@total' => $sandbox['total'],
--- a/modules/islandora_scholar_embargo/includes/batch.inc
+++ b/modules/islandora_scholar_embargo/includes/batch.inc
@@ -155,32 +155,45 @@ function islandora_scholar_embargo_reapply_embargoes_batch_operation(&$context) 
   module_load_include('inc', 'islandora_scholar_embargo', 'includes/embargo');
   $sandbox =& $context['sandbox'];
   if (empty($sandbox)) {
-    $sandbox['total'] = islandora_scholar_embargo_get_all_embargoed(TRUE);
     $sandbox['completed'] = 0;
+    $sandbox['offset'] = '1900-01-01T00:00:00.000Z';
   }
 
-  // Iterate through a set.
-  $set = islandora_scholar_embargo_get_embargoed_items_subset($sandbox['completed']);
-  if (!empty($set)) {
-    foreach ($set as $embargoed_item) {
-      $id_parts = explode('/', $embargoed_item['obj']['value']);
-      $pid = array_shift($id_parts);
-      $dsid = array_shift($id_parts);
-      $dsids = is_null($dsid) ? $dsid : array($dsid);
-      islandora_scholar_embargo_lift_embargo($pid, $dsids);
-      islandora_scholar_embargo_set_embargo($pid, $dsids, $embargoed_item['date']['value']);
-      $sandbox['completed']++;
-      $context['message'] = t('Re-applied embargo to @item of @pid.', array(
-        '@item' => is_null($dsid) ? t('entire object') : $dsid,
-        '@pid' => $pid,
-      ));
+  // It's possible for this number to change during the batch (e.g., if an
+  // embargo expires or a datastream is modified), so re-grabbing is safest to
+  // prevent items from being skipped.
+  $sandbox['total'] = islandora_scholar_embargo_get_all_embargoed(TRUE);
+  if (!$sandbox['total']) {
+    $context['message'] = t('No items to embargo found, or embargoed items set was reduced to zero at some point during the batch.');
+    return;
+  }
+  else {
+    // Iterate through a set.
+    $set = islandora_scholar_embargo_get_embargoed_items_subset($sandbox['offset']);
+    if (!empty($set)) {
+      foreach ($set as $embargoed_item) {
+        $id_parts = explode('/', $embargoed_item['obj']['value']);
+        $pid = array_shift($id_parts);
+        $dsid = array_shift($id_parts);
+        $dsids = is_null($dsid) ? $dsid : array($dsid);
+        islandora_scholar_embargo_lift_embargo($pid, $dsids);
+        islandora_scholar_embargo_set_embargo($pid, $dsids, $embargoed_item['expiry']['value']);
+        $sandbox['completed']++;
+        $sandbox['offset'] = $embargoed_item['date']['value'];
+        $context['message'] = t('@count/@total: Re-applied embargo to @item of @pid.', array(
+          '@count' => $sandbox['completed'],
+          '@total' => $sandbox['total'],
+          '@item' => is_null($dsid) ? t('entire object') : $dsid,
+          '@pid' => $pid,
+        ));
+      }
+      // If the set was empty, $context['finished'] will be 1 anyway.
+      $context['finished'] == $sandbox['completed'] / $sandbox['total'];
     }
-  }
-
-  $context['finished'] = $sandbox['completed'] / $sandbox['total'];
-  if ($context['finished'] == 1) {
-    drupal_set_message(t('Re-applied embargoes to @total objects.', array(
-      '@total' => $sandbox['completed'],
-    )));
+    if ($context['finished'] >= 1) {
+      drupal_set_message(t('Re-applied embargoes to @total objects.', array(
+        '@total' => $sandbox['completed'],
+      )));
+    }
   }
 }
--- a/modules/islandora_scholar_embargo/includes/batch.inc
+++ b/modules/islandora_scholar_embargo/includes/batch.inc
@@ -125,3 +125,62 @@ function islandora_scholar_embargo_lift_batch_finished($success, $results, $oper
     drupal_set_message(format_plural($results['skip_count'], '@count object skipped as the content model is not configured to allow embargoes.', '@count objects skipped as their content models are not configured to allow embargoes.'));
   }
 }
+
+/**
+ * Batch definition for re-applying embargos to embargoed objects.
+ *
+ * @return array
+ *   The batch definition.
+ */
+function islandora_scholar_embargo_reapply_embargos_batch() {
+  return array(
+    'title' => t('Re-Apply Embargos to Embargoed Objects'),
+    'error_message' => t('An error has occurred and the embargos may not have all been re-applied.'),
+    'init_message' => t('Preparing to re-apply embargos to embargoed objects.'),
+    'progress_message' => t('Time elapsed: @elapsed <br/>Estimated time remaining: @estimate.'),
+    'file' => drupal_get_path('module', 'islandora_scholar_embargo') . '/includes/batch.inc',
+    'operations' => array(
+      array('islandora_scholar_embargo_reapply_embargos_batch_operation', array()),
+    ),
+  );
+}
+
+/**
+ * Batch operation for re-applying embargos to embargoed objects.
+ *
+ * @param array $context
+ *   The batch context.
+ */
+function islandora_scholar_embargo_reapply_embargos_batch_operation(&$context) {
+  module_load_include('inc', 'islandora_scholar_embargo', 'includes/embargo');
+  $sandbox =& $context['sandbox'];
+  if (empty($sandbox)) {
+    $sandbox['total'] = islandora_scholar_embargo_get_all_embargoed(TRUE);
+    $sandbox['completed'] = 0;
+  }
+
+  // Iterate through a set.
+  $set = islandora_scholar_embargo_get_embargoed_items_subset($sandbox['completed']);
+  if (!empty($set)) {
+    foreach ($set as $embargoed_item) {
+      $id_parts = explode('/', $embargoed_item['obj']['value']);
+      $pid = array_shift($id_parts);
+      $dsid = array_shift($id_parts);
+      $dsids = is_null($dsid) ? $dsid : array($dsid);
+      islandora_scholar_embargo_lift_embargo($pid, $dsids);
+      islandora_scholar_embargo_set_embargo($pid, $dsids, $embargoed_item['date']['value']);
+      $sandbox['completed']++;
+      $context['message'] = t('Re-applied embargo to @item of @pid.', array(
+        '@item' => is_null($dsid) ? t('entire object') : $dsid,
+        '@pid' => $pid,
+      ));
+    }
+  }
+
+  $context['finished'] = $sandbox['completed'] / $sandbox['total'];
+  if ($context['finished'] == 1) {
+    drupal_set_message(t('Re-applied embargos to @total objects.', array(
+      '@total' => $sandbox['completed'],
+    )));
+  }
+}
--- a/modules/islandora_scholar_embargo/includes/batch.inc
+++ b/modules/islandora_scholar_embargo/includes/batch.inc
@@ -127,31 +127,31 @@ function islandora_scholar_embargo_lift_batch_finished($success, $results, $oper
 }
 
 /**
- * Batch definition for re-applying embargos to embargoed objects.
+ * Batch definition for re-applying embargoes to embargoed objects.
  *
  * @return array
  *   The batch definition.
  */
-function islandora_scholar_embargo_reapply_embargos_batch() {
+function islandora_scholar_embargo_reapply_embargoes_batch() {
   return array(
     'title' => t('Re-Apply Embargos to Embargoed Objects'),
-    'error_message' => t('An error has occurred and the embargos may not have all been re-applied.'),
-    'init_message' => t('Preparing to re-apply embargos to embargoed objects.'),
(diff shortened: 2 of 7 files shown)


modules/islandora_scholar_embargo/includes/admin.form.inc
```diff
@@ -21,7 +21,9 @@ function islandora_embargo_admin(array $form, array &$form_state) {
   $options = islandora_get_content_models();
   $selected = variable_get('islandora_embargo_content_models', array('ir:citationCModel', 'ir:thesisCModel'));
   foreach ($selected as $cmodel) {
-    $options = array($cmodel => $options[$cmodel]) + $options;
+    if (isset($options[$cmodel])) {
+      $options = array($cmodel => $options[$cmodel]) + $options;
+    }
```
adam-vessey, line 26: I'm... not sure I understand what's going on here. If there's an entry in `$options`, we reset its value to itself?

Are we looking for something more like:

``` php
$options = array_intersect_key($options, drupal_map_assoc($selected));
```

instead of this `foreach` over `$selected`?

... Actually... This was some craziness to establish the "desired" ordering or something, was it?

qadan, line 26: It's just because it's possible for those options not to be returned in `islandora_get_content_models()` due to XACML or other what-have-ye's, so `$options[$cmodel]` may not exist. But yeah, there's probably a more PHP-weird-array-business way to do it?

adam-vessey, line 26: Yeah, it wasn't so much the `isset()`, but the whole: "options = one of it's items + the rest of it"...  Which would only have the affect of (re)ordering the array...?

qadan, line 26: Perhaps the intention was to place the current config at the top? Because ... UI design? I dunno?

modules/islandora_scholar_embargo/includes/embargo.inc
```diff
@@ -445,24 +445,60 @@ EOQ;
...
+ *   An offset to start from.
+ * @param int $limit
+ *   The number of items to return in the subset.
+ *
+ * @return array
+ *   An array containing a subset of embargoed items.
+ */
+function islandora_scholar_embargo_get_embargoed_items_subset($offset = 0, $limit = 10) {
+  $current_embargo = islandora_scholar_embargo_get_embargoed_items_query() . <<<EOQ
+OFFSET $offset
```
adam-vessey: Should generally avoid slicing by numerical offsets... Say the embargo for an object expires (and has the relationship removed) while processing: We could end up skipping objects. Order ascending by created date and PID, and offset by created date, breaking ties with PIDs when necessary.

modules/islandora_scholar_embargo/includes/embargo_roles_manage.inc
```diff
@@ -0,0 +1,75 @@
...
+ * @param array $form_state
+ *   The state of the submitted form.
+ */
+function islandora_scholar_embargo_manage_roles_admin_form_submit($form, &$form_state) {
+  // Attach role names to role IDs to make it more useful as a variable.
+  $whitelist = array();
+  foreach ($form_state['values']['islandora_scholar_embargo_whitelisted_roles'] as $rid) {
+    $whitelist[$rid] = $form['roles_wrapper']['islandora_scholar_embargo_whitelisted_roles']['#options'][$rid];
+  }
+  variable_set('islandora_scholar_embargo_whitelisted_roles', $whitelist);
```
adam-vessey, line 68: Missing `variable_del()` for the new variable?

qadan, line 68: whooops

modules/islandora_scholar_embargo/includes/embargo_roles_manage.inc
```diff
@@ -0,0 +1,75 @@
...
+        '#default_value' => array_keys(variable_get('islandora_scholar_embargo_whitelisted_roles', array())),
+      ),
+    ),
+    'note' => array(
+      '#markup' => t('Saving and applying will initiate a batch to modify the XACML policies of all embargoed objects. Bear in mind that this can take a long time to execute depending on the number of items to be embargoed, and that it may be more prudent to run the Drush command <code>islandora_scholar_embargo_apply_roles_to_embargoed_objects</code> instead. Number of objects to be embargoed: <strong>@number</strong>', array(
+        '@number' => number_format(islandora_scholar_embargo_get_all_embargoed(TRUE)),
+      )),
+    ),
+    'submit_buttons' => array(
+      '#type' => 'item',
```
adam-vessey: [`actions` is a thing](https://api.drupal.org/api/drupal/developer!topics!forms_api_reference.html/7.x#actions).

modules/islandora_scholar_embargo/includes/embargo_roles_manage.inc
```diff
@@ -0,0 +1,75 @@
...
+      '#collapsible' => FALSE,
+      'islandora_scholar_embargo_whitelisted_roles' => array(
+        '#type' => 'checkboxes',
+        '#description' => t('Check off all the roles that should have access to embargoed items. <strong>NOTE:</strong> since embargo XACML role entries are whitelists, embargoed objects in a Fedora repository shared between Drupal sites will be visible to the same role in both sites. Consider using site-unique roles when dealing with shared repositories.'),
+        '#options' => user_roles(TRUE, ISLANDORA_SCHOLAR_EMBARGO_CAN_EMBARGO_ANY),
+        '#default_value' => array_keys(variable_get('islandora_scholar_embargo_whitelisted_roles', array())),
+      ),
+    ),
+    'note' => array(
+      '#markup' => t('Saving and applying will initiate a batch to modify the XACML policies of all embargoed objects. Bear in mind that this can take a long time to execute depending on the number of items to be embargoed, and that it may be more prudent to run the Drush command <code>islandora_scholar_embargo_apply_roles_to_embargoed_objects</code> instead. Number of objects to be embargoed: <strong>@number</strong>', array(
```
adam-vessey: Would it be better to indicate that the number is approximate? Thinking users could get spooked if they see a number here, but then it processes more or fewer in actuality.

qadan: "Approx. 15,217 objects to be embargoed" might sound funny but I think it can be worded to make sense haha

modules/islandora_scholar_embargo/includes/batch.inc
```diff
@@ -125,3 +125,62 @@ function islandora_scholar_embargo_lift_batch_finished($success, $results, $oper
...
+function islandora_scholar_embargo_reapply_embargoes_batch_operation(&$context) {
+  module_load_include('inc', 'islandora_scholar_embargo', 'includes/embargo');
+  $sandbox =& $context['sandbox'];
+  if (empty($sandbox)) {
+    $sandbox['total'] = islandora_scholar_embargo_get_all_embargoed(TRUE);
+    $sandbox['completed'] = 0;
+  }
+
+  // Iterate through a set.
+  $set = islandora_scholar_embargo_get_embargoed_items_subset($sandbox['completed']);
```
adam-vessey: Could potentially loop forever here, if things get deleted.

qadan: I've got an alteration to this that should fix 'er

modules/islandora_scholar_embargo/islandora_scholar_embargo.module
```diff
@@ -474,7 +483,10 @@ function islandora_scholar_embargo_set_embargo($param, $dsid = NULL, $end = 'ind
       }
       foreach ($users as $user) {
         $xacml->viewingRule->addUser($user->name);
-        $xacml_changed = TRUE;
+      }
+      foreach ($roles as $role) {
+        $xacml->viewingRule->addRole($role);
+        $xacml->managementRule->addRole($role);
```
adam-vessey: Is management really necessary? How could others manage something they can't see?

modules/islandora_scholar_embargo/includes/admin.form.inc
adam-vessey, line 26: Shrug. Is fine.

modules/islandora_scholar_embargo/includes/embargo.inc
```diff
@@ -445,24 +445,63 @@ EOQ;
...
+PREFIX is: <info:islandora/islandora-system:def/scholar#>
+SELECT DISTINCT ?obj ?expiry ?date
+FROM <#ri>
+WHERE {
+  ?obj is:embargo-until ?expiry ;
+  { ?obj <fedora-model:createdDate> ?date } UNION
+  { ?obj <fedora-view:lastModifiedDate> ?date ;
+         <fedora-view:disseminationType> ?dsid
+  }
+  FILTER(?date > "$offset"^^<http://www.w3.org/2001/XMLSchema#dateTime>)
```
adam-vessey: Could miss items, two (or more) items happen to have the same date (and the set crosses a slice boundary). Would have to be something like "date greater or (date equal and something to break the tie)"... Usually using the PID...

``` sparql
PREFIX is: <info:islandora/islandora-system:def/scholar#>
SELECT DISTINCT ?obj ?expiry ?date
FROM <#ri>
WHERE {
  ?obj is:embargo-until ?expiry ;
  { ?obj <fedora-model:createdDate> ?date } UNION
  { ?obj <fedora-view:lastModifiedDate> ?date ;
         <fedora-view:disseminationType> ?dsid
  }
  FILTER(?date > "$offset"^^<http://www.w3.org/2001/XMLSchema#dateTime> || (?date = "$offset"^^<http://www.w3.org/2001/XMLSchema#dateTime> && ?obj > "$the_last_one"))
}
ORDER BY ASC(?date), ASC(?obj)
```

or something of the like.

modules/islandora_scholar_embargo/includes/embargo_roles_manage.inc
adam-vessey, line 68: Also, still?

qadan:
It'll likely be the former; I don't think #220 will correctly cherry-pick into this one. I was going to wait on resolution of this ticket before taking care of the ticket cleanup

qadan:
@ruebot yep; that conversation actually happened in https://jira.duraspace.org/browse/ISLANDORA-1602! I asked him if I could take over the work before starting on this.

qadan:
That was indeed the original plan, but I attempted to indicate in the [last comment on there](https://jira.duraspace.org/browse/ISLANDORA-1602?focusedCommentId=49440&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-49440) that the plan changed due to a change in the method taken and the number of things that had to be altered; sorry if that wasn't clear.

modules/islandora_scholar_embargo/includes/embargo.inc
```diff
@@ -485,7 +488,16 @@ EOQ;
...
  *   as the item's created or last modified date.
  */
-function islandora_scholar_embargo_get_embargoed_items_subset($offset = '1900-01-01T00:00:00.000Z', $limit = 10) {
+function islandora_scholar_embargo_get_embargoed_items_subset($offset_date = '1900-01-01T00:00:00.000Z', $offset_pid = NULL, $limit = 10) {
+  // Establish some parameters for the query.
+  $filter = is_null($offset_pid) ?
+    "FILTER(?date > \"$offset_date\"^^<http://www.w3.org/2001/XMLSchema#dateTime>)" :
+    "FILTER(?date > \"$offset_date\"^^<http://www.w3.org/2001/XMLSchema#dateTime> || (?date = \"$offset_date\"^^<http://www.w3.org/2001/XMLSchema#dateTime> && str(?obj) > \"$offset_pid\"))";
+  $order_by = is_null($offset_pid) ?
+    "ORDER BY ASC(?date)" :
```
adam-vessey: This would still have to order `ASC(?obj)`, to ensure we can pass correctly to the second slice. As in, still: `ORDER BY ASC(?date), ASC(?obj)`.

qadan: Thought could save some time on the ordering, but that makes sense. Repair'd

modules/islandora_scholar_embargo/includes/embargo.inc
```diff
@@ -485,7 +488,16 @@ EOQ;
  *   including 'obj' as the object ID, 'expiry' as the expiry date, and 'date'
  *   as the item's created or last modified date.
  */
-function islandora_scholar_embargo_get_embargoed_items_subset($offset = '1900-01-01T00:00:00.000Z', $limit = 10) {
+function islandora_scholar_embargo_get_embargoed_items_subset($offset_date = '1900-01-01T00:00:00.000Z', $offset_pid = NULL, $limit = 10) {
+  // Establish some parameters for the query.
+  $filter = is_null($offset_pid) ?
+    "FILTER(?date > \"$offset_date\"^^<http://www.w3.org/2001/XMLSchema#dateTime>)" :
```
adam-vessey: Could we not just avoid specifying the filter entirely, until we have a value based on data? Just use the empty string, and have things be in the proper order. We'll grab the first slice of `$limit` items, and next time we get called, we'll have the two values at which to start the next slice.

modules/islandora_scholar_embargo/includes/embargo_roles_manage.inc
adam-vessey, line 68: Fixed in https://github.com/Islandora/islandora_scholar/pull/232/commits/420aa124a087d663d20702cfda37e707c0f5121f

qadan:
@ruebot I honestly had not realized that it was listed as a bug! I'd assumed throughout that this was an improvement. I don't know, personally, if I'd consider this to be a bug? Looking at the former discussion on this, it seems to waver somewhere between "is a bug" and "is working as intended but is misleading"

qadan:
Definitely will bring up; there's either two or three 7.x-1.7 pulls left to make depending on how this would be considered.

modules/islandora_scholar_embargo/islandora_scholar_embargo.module
```diff
@@ -502,14 +513,15 @@ function islandora_scholar_embargo_set_embargo($param, $dsid = NULL, $end = 'ind
           $embargo_ds->relationships->add(ISLANDORA_SCHOLAR_EMBARGO_RELS_URI, ISLANDORA_SCHOLAR_EMBARGO_NOTIFICATION_PRED, $notification, $type);
         }
         $xacml->datastreamRule->addDsid($datastream_id);
-        $xacml_changed = TRUE;
         foreach ($users as $user) {
```
nigelgbanks, line 516: We need another `foreach` here to apply the roles as is done above.

``` php
foreach ($roles as $role) {
    $xacml->datastreamRule->addRole($role);
}
```

nigelgbanks:
Looks like we need some additional work? Newly embargoed objects don't inherit the whitelist. This only applies when embargoing data-streams, embargoing the whole object works.

# Steps to reproduce.
1. Create new role 'test'.
2. Add permission 'Manage embargo on any objects'.
3. Add new role 'test' to the whitelist here `admin/islandora/solution_pack_config/embargo/roles`
4. Embargo and existing object that currently does not have an embargo set (You **must** choose data-steam based embargo).
5. View the XACML data-stream and note the permissions.

See the comment here for a fix: https://github.com/Islandora/islandora_scholar/pull/232/files#r64203012

@qadan I've sent you a pull with the roles added.

https://github.com/qadan/islandora_scholar/pull/1/files

qadan:
@ruebot true; that is covered under 5.iii of the committers workflow. Would it have made more sense to reject the PR and implement the line myself? I'm personally fine with it as-is, especially given the 'PR-as-code-review-recommendation' nature of the commit, and given the discussion surrounding this ticket on the last committers call regarding the general unfamiliarity with Scholar; @nigelgbanks was the only one the group could find outside of DGI willing to do the review/merge

manez:
+1 for this being within the spirit of the law, if not quite the letter. As long as we've got clear communication around what's happening (and I think we do here), I don't see the benefit in backing this one out and redoing it. 

While noting that yes, it was technically a self-commit and shouldn't have gone in this way.

bryjbrown:
Looks like its a bit too late to be helpful since the pull has already been merged, but I did complete the smoke test for this pull on an Islandora Vagrant machine today. To summarize [the more fleshed out comment in JIRA](https://jira.duraspace.org/browse/ISLANDORA-1717?focusedCommentId=50168&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-50168), it works but seems to throw a nontrivial amount of errors and the secondary layer of permissions created by the whitelist might be a bit confusing to users (although links to the whitelist in the permission description do help). The batch embargo updater is great when it works, but seems to throw a 500 AJAX error roughly half the time.

whikloj:
@bryjbrown, I'd suggest a new JIRA ticket to fix this new feature and get it into a working state.

bryjbrown:
@whikloj Someone else is free to carry the torch on this if they like, otherwise I'm just going to let it be. 
I'm not very heartened by the path this has taken (PR rejected, obvious bug changed to a new feature request in JIRA, completely new solution designed without input from community, etc). 

We (FSU/FLVC) resolved the issue to our own satisfaction with the pull request included in [my original ticket (JIRA 1602)|https://jira.duraspace.org/browse/ISLANDORA-1602], and are just going to stay on our FLVC fork of Scholar for the time being instead of updating. What's left of this issue should be pushed forward by someone with more skin in the game.

Also, apologies if this comes across as rude to anyone, particularly anyone who made commits on this. I appreciate the work thats gone into this & thank you for it, and I'll take the blame on this for stalling on the ticket when it was assigned to me back in April and misunderstanding the process that followed. I'll try to be more active in the process next time to scratch my own itches. 😄

qadan:
@bryjbrown Not rude, I think! It's probably far more healthy to get your concerns out in the open rather than otherwise.

To explain the decision behind the whitelist and the dual-permissions thing: one deals with Drupal permisssions and the other deals with XACML permissions. While I agree it mostly seems like a secondary layer, precedent so far has exclusively leaned towards not using Drupal permissions to dictate XACML permissions, and I didn't really see a pressing enough need to deviate from that precedent.

The bug-to-feature change was done on a committer's call; I think it boiled down to the above precedent again - Drupal permissions don't currently dictate XACML permissions anywhere, and instead so far we've opted to document this fact on the Drupal permissions page anywhere XACML could override Drupal, so the documentation was updated.

As to the last point, though ... I'm honestly not sure what the better path was here for input from the community? This pull was opened April 20th and closed over a month later, I believe? It was brought up a couple times in committer's, and I pinged the interested parties on here at least a couple times before it got in as well, but I just don't know what else I can do beyond these types of things. The last committer's call this was brought up in, I couldn't find anyone willing to review it besides @nigelgbanks (which, thanks Nigel!) - somewhat understandable, since this is pretty niche functionality, but honestly more input would be fantastic! Especially when there's concerns like the one with the whitelist. Just ... I dunno, over a month later, well past its original review, I'm not sure what other means are available to me.

If you feel like the original PR was squashed, that's my bad; I think muscle memory just makes me "WRITE CODE MAKE PR" haha, and I probably could have PR'd my additions against yours. I wish I'd known you'd felt as such; I could have just closed this one and done that instead.

And as for the bugs found, I'll take a peek soon! If I've submitted something broken, I should probably clean up my mess >_>

bryjbrown:
@qadan That's good to know that the reason for the whitelist had to do with staying consistent with a previous precedent. While I don't necessarily agree with the decision (but would be open to discussion about it), I can at least appreciate the logic behind it.

I think in the future though, we should try to keep all discussion of the problem under one JIRA ticket to make it linear and easy to follow. The jump from [JIRA 1602](https://jira.duraspace.org/browse/ISLANDORA-1602) to [JIRA 1717](https://jira.duraspace.org/browse/ISLANDORA-1717) is never made explicit, and this is confusing for people on the outside looking in (such as @DonRichards in [this comment](https://jira.duraspace.org/browse/ISLANDORA-1602?focusedCommentId=50190&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-50190) and trying to follow the thread from bug description through discussion of design and implementation. Having to skip back and forth between multiple tickets adds more complexity and should be avoided if possible. 

All in all I think I was just having a rough afternoon yesterday and was already feeling pretty salty, and @whikloj's suggestion about creating a third ticket prompted me to post a fairly tactless comment (I know this was not his intention of course). I appreciate @qadan addressing my concerns about this issue as well, hearing his side of the story puts things in a better perspective. 👍

qadan:
@bryjbrown definitely agree; this one got a little ... weird. I think a lot of that's on me. I'll try to keep things nice and clean in the future. And I'll see about those errors too; seems kinda troubling

PR #233, opened 2016-05-09
Whitelist instead of blacklist.

jordandukart:
**JIRA Ticket**: https://jira.duraspace.org/browse/ISLANDORA-1714

Whitelists DOI support versus blacklisting. Currently we only support journals and journal articles.
# How should this be tested?

Import the DOI 10.1117/12.2197868
Note that PHP whitescreens with a SimpleXML error.
After pulling the code no longer whitescreens and a warning is displayed instead.
# Interested parties

@Islandora/7-x-1-x-committers

### DIFF
--- a/modules/doi/modules/doi_importer/includes/importer.inc
+++ b/modules/doi/modules/doi_importer/includes/importer.inc
@@ -129,7 +129,11 @@ class DOIImportObject extends IslandoraImportObject {
       $crossref_xml = simplexml_load_string($response->data);
       // Totally don't support books so let's not attempt to make records that
       // are actually books!
-      if ($crossref_xml->doi_record->crossref->error || $crossref_xml->doi_record->crossref->book) {
+      if ($crossref_xml->doi_record->crossref->error) {
+        return FALSE;
+      }
+      elseif (!$crossref_xml->doi_record->crossref->journal) {
+        drupal_set_message(t("Only the 'Journal' type is currently supported for DOI importing."), 'error');
         return FALSE;
       }
       else {


mjordan:
I'm likely testing this wrong. I can't reproduce the WSOD but I'm getting an empty batch ingest set. Working off islandora_scholar 00a40da17fc98ffd22c456b071e9cdbee3aa520f.

Steps I followed:
1. In my collection's Manage tab, click on "Batch import objects"
2. Choose "DOI importer" from select list, click on "Next"
3. Paste '10.1117/12.2197868' into the text area, click on "Import"
4. Batch runs, no WSOD, green "Processing complete" confirmation message tells me to check the queue.
5. When I check the queue, the batch set has 0 members.

Is this a different problem or am I not understanding how to test properly?

jordandukart:
Did you enter a valid CrossRef account in the configuration for the DOI Importer?

qadan:
screen shot 2016-05-20 at 9 43 37 am

This is what I get, which I believe is the intended behaviour? Just not failing horribly when adding a non-journal DOI.

mjordan:
Now that I've registered with CrossRef and configured Scholar, I get what Dan shows above. But, even with a configured CrossRef account, while still running on the 7.x code (before switching to Jordan's branch), I could import the DOI without getting a WSOD. So I was able to confirm the fix but not the original problem.

nigelgbanks:
Hmm seems like this is the DOI that is problematic? `10.1117/12.2197769` perhaps try it instead of `10.1117/12.2197868` and you'll be able to reproduce the problem on 7.x?

mjordan:
@nigelgbanks yup, using that DOI on 7.x I can replicate the error:

```
[Fri May 20 18:28:23.827969 2016] [:error] [pid 28410] [client 10.0.2.2:33044] PHP Fatal error:  Call to a member function asXML() on a non-object in /var/www/drupal/sites/all/modules/islandora_scholar/modules/doi/modules/doi_importer/includes/importer.inc on line 154, referer: http://localhost:8000/batch?op=start&id=463&render=overlay
```

And on Jordan's branch I get the user-facing warning pictured above.

So I'm :+1: to merge this unless this smoke test was too convoluted. Let me know if you think it was.

nigelgbanks:
I think your good to merge.

PR #236, opened 2016-06-24
api documentation

qadan:
**JIRA Ticket**: https://jira.duraspace.org/browse/ISLANDORA-1729

Just adds some API documentation for Scholar Embargo.
# What's new?

A newly-documented (but previously-existing) alter hook in embargo.

Also some sample code in there.
# How should this be tested?

Suppose you could try implementing the hooks and seeing if the sample code does what it says on the tin, but realistically it's just documentation.
# Interested parties

Uh ... not sure; I'm the maintainer of the module, I brought it up, I'm sending it to completion. IR-types, maybe?

### DIFF
--- a/modules/islandora_scholar_embargo/islandora_scholar_embargo.api.php
+++ b/modules/islandora_scholar_embargo/islandora_scholar_embargo.api.php
@@ -37,7 +37,7 @@ function hook_islandora_scholar_embargo_users_to_notify_alter(array &$users, Abs
   // Example: don't notify the anonymous user if the object is a citation.
   if (in_array('ir:citationCModel', $object->models)) {
     $filter_anonymous = function($user) {
-      return user->name !== 'anonymous';
+      return $user->name !== 'anonymous';
     };
     $users = array_filter($users, $filter_anonymous);
   }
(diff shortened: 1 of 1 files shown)


DonRichards:
I think saying this is adds API documentation might be misleading. When I think of documentation I think either markdown or code comments. This adds functionality. But I do understand why you said that because it informs the user.

DonRichards:
There seems to be an unrelated issue with Travis. Looking into it now

qadan:
Silly syntax issue; cleaned it up.

The .api.php is sort of kind of technically documentation in Drupal's weird way? It's supposed to be parsed by the API module and transformed into hook documentation for http://api.drupal.org - it's not actually loaded by Drupal or intended to be functional (re: https://www.drupal.org/coding-standards/docs#hooks). In an ideal world we'd be Doxygen-izing our modules' .api.phps and uploading the results somewhere.

DonRichards:
Now that this is linting fine. How do I test the API call? Step?

qadan:
Hmm ... you just have to implement it, I guess. A test/stub module that implements `hook_islandora_scholar_embargo_users_to_notify_alter()` should be able to see what are in the `$users` and `$object` variables, and should be able to modify the `$users` array by reference. Other than that ... not much to test by way of the UI

bryjbrown (approved):
Tested with a stub module that only implements these hooks, and it works as advertised. Thanks for the clear documentation @qadan and apologies for the delay in merging!

qadan:
No worries! Thanks for the merge 👍

PR #250, opened 2017-02-03
ISLANDORA-1100: Fixes syntax in ISSN counting logic

bryjbrown:
**JIRA Ticket**: [ISLANDORA-1100](https://jira.duraspace.org/browse/ISLANDORA-1100)

Fixes the ISSN counting syntax in [RomeoView::getISSN()](https://github.com/Islandora/islandora_scholar/blob/7.x/includes/romeo.tab.inc#L31-L68) that was causing the following error message:
RomeoView::getISSN() error message

The error message shown in [ISLANDORA-1100](https://jira.duraspace.org/browse/ISLANDORA-1100) happens when an object with no `<identifier type="issn">` is loaded with 'Enable RoMEO attempts.' checked in the Scholar config (admin/islandora/solution_pack_config/scholar). It only happens when the ISSN isn't cached, so it can be reproduced by ingesting a new object or manually clearing Drupal's cache (but will only happen on the first page load, after running RomeoView::getISSN() once it will be cached and no longer produce the error).

When Scholar tries to get the ISSN from a record, [line 55](https://github.com/Islandora/islandora_scholar/blob/7.x/includes/romeo.tab.inc#L55) incorrectly tries to determine if there was a match or not. The current behavior runs `(count($issn_results) > 0)` to check to see if the xpath query got at least one match, but this expression will always return true because the xpath result is a DOMNodeList object, not an array. Since it isnt an array, a count on it returns 1 and it moves on to the next line where its trying to get the nonexistent value from the result, hence the error message.

The replacement line, `($issn_results->length > 0)`, should get the proper number of results so that the next line only fires when there really are no ISSNs.

# What's new?
Simple change in an if statement on [line 55](https://github.com/bryjbrown/islandora_scholar/blob/ISLANDORA-1100/includes/romeo.tab.inc#L55) of islandora_scholar/includes/romeo.tab.inc from `(count($issn_results) > 0)` to `($issn_results->length > 0)`.

# How should this be tested?
Enable Romeo attempts from the Scholar config screen and then load a new object with no `<identifier type='issn'>`. On the current branch, you should see an error message from RomeoView::getISSN() on the very first view (subsequent views won't trigger it because its cached), and you can trigger the error again by clearing the cache and re-viewing the object.

Switching to the ISLANDORA-1100 branch, clearing the cache and re-viewing the object should not trigger an error.

# Interested parties
@Islandora/7-x-1-x-committers, @DonRichards (co-maintainer), @dmoses (error reporter)

### DIFF
--- a/includes/romeo.tab.inc
+++ b/includes/romeo.tab.inc
@@ -52,7 +52,7 @@ class RomeoView {
           $mods_xpath->registerNamespace('m', 'http://www.loc.gov/mods/v3');
 
           $issn_results = $mods_xpath->query('//m:identifier[@type="issn" and normalize-space(text())]');
-          if (count($issn_results) > 0) {
+          if ($issn_results->length > 0) {
             $issn = $issn_results->item(0)->textContent;
           }
         }


DonRichards:
OK, I'll see if I can test this today. I have training for the rest of the week so if not today it'll be Monday.

DonRichards:
## I'm having an issue replicating the problem.

### Without the pull:


### With the pull:


Just to be clear, my steps so I can see what I'm missing to replicate the error. 
1. Vagrant destroy && vagrant up 
2. Logged in as admin
3. Went to admin/islandora/solution_pack_config/scholar
4. Added Sherpa/RoMEO API Key value only > Save
5. Navigation > Islandora Repository > Citations > Manage > Add an object to this Collection
6. Citation Content model > Skipped MARCXML
7. Entered Title, First Name, Last name, Qualified Name, Date Issued
8. Checked "Would you like to include a PDF document for this citation?"
9. Uploaded a file
10. Published, Contact Author (I hold the copyright and wish to retain all rights)., I certify > Ingest

bryjbrown:
@DonRichards Sometimes the GUI metadata entry form enters empty elements which may be messing with this. Try to verify that there is no identifier@type="issn" element present in the metadata (not even an empty one) and then run `drush cc all` and then refresh the view of the object. I get an error message.

bryjbrown:
@DonRichards I also don't enter a Sherpa/RoMEO API key (leave it blank) and check the "Enable RoMEO attempts" checkbox and save. That last part is important, it won't execute the code in the PR without that being turned on.

DonRichards:
I'm being lazy. What command are you using?

bryjbrown:
@DonRichards To clear the cache? `drush cc all` from the CLI, or click "Clear all caches" from /admin/config/development/performance.

DonRichards:
lol, sorry I meant your drush command to ingest to bypass the gui adding in blank fields.

bryjbrown:
@DonRichards Ah, I just used the zip importer for a MODS record I had handy w/o ISSN. 

I just tested it using the GUI form and even though it added an empty `<identifier type="issn"/>`, it still triggered the RomeoView::getISSN() error after I cleared the cache. 

Here's what my SherpaRoMEO config looks like: 
config
Notice that it is completely stock with the exception of "Enable RoMEO attempts" being checked.

And here's the error I get from an ISSN-less record after a `drush cc all`:
error

jordandukart (approved):
Removes a long standing warning.

DiegoPino:
thanks @jordandukart !!

PR #255, opened 2017-02-27
ISLANDORA-1098: Adds var check before loading DOI importer form to see if OpenURL PID is configured properly

bryjbrown:
**JIRA Ticket**: [ISLANDORA-1098](https://jira.duraspace.org/browse/ISLANDORA-1098)

Blocks users from using the DOI Importer form when they haven't configured the OpenURL PID email address.

# What's new?
Adds a check for the Drupal variable 'islandora_doi_openurl_pid' to the top of [IslandoraBatchImporter::getForm()](https://github.com/Islandora/islandora_scholar/blob/7.x/modules/doi/modules/doi_importer/includes/importer.inc#L22-L59). If the variable hasn't been set and the default 'user@example.com' is being used, return an error message w/ link to admin page telling them they need to configure an email address instead of loading the DOI Importer form.

The reason I chose to block loading of the DOI Importer form is that the form will silently fail if a bad email address is used, and due to the way the importer framework is implemented, I cant add a validation function to run after the DOI Importer form is submitted. Better to totally block loading of the form with a helpful error message instead.

# How should this be tested?
Set the "OpenURL PID" value in /admin/islandora/solution_pack_config/scholar/islandora_doi to the default, 'user@example.com', then try to import objects using the DOI Importer (/islandora/object/$collection_pid/manage/collection/batch_import). it should show the following error:
screen shot 2017-02-27 at 10 43 56 am

Now set the "OpenURL PID" value to any other email address other than the default and try to use the DOI Importer again, it should load normally.

# Additional Notes:
I never ingest objects via DOI, so the DOI Importer page is the only one I know of where I could insert a check to stop the user from importing via the default OpenURL PID. If there are other places where I should add warnings let me know.

# Interested parties
@DiegoPino @dmoses @DonRichards @willtp87 @jordandukart (pinging last two because of related ticket, [ISLANDORA-1291](https://jira.duraspace.org/browse/ISLANDORA-1291))

### DIFF
--- a/modules/doi/modules/doi_importer/includes/importer.inc
+++ b/modules/doi/modules/doi_importer/includes/importer.inc
@@ -27,7 +27,7 @@ class DOIImporter extends IslandoraBatchImporter {
   public static function getForm(array & $form_state) {
     $openurl_pid = variable_get('islandora_doi_openurl_pid', 'user@example.com');
     if ($openurl_pid == 'user@example.com') {
-      drupal_set_message(t('OpenURL PID email address has not been configured, please enter a valid email address in the <a href="@doi-admin-page">Scholar DOI configuration form</a>.', array('@doi-admin-page' => '/admin/islandora/solution_pack_config/scholar/islandora_doi',)), 'error');
+      drupal_set_message(t('OpenURL PID email address has not been configured, please enter a valid email address in the <a href="@doi-admin-page">Scholar DOI configuration form</a>.', array('@doi-admin-page' => '/admin/islandora/solution_pack_config/scholar/islandora_doi')), 'error');
     }
     else {
       return array(
--- a/modules/doi/modules/doi_importer/includes/importer.inc
+++ b/modules/doi/modules/doi_importer/includes/importer.inc
@@ -4,6 +4,7 @@
  * @file
  * Defines a class for importing DOI citations.
  */
+
 class DOIImporter extends IslandoraBatchImporter {
 
   protected $itemClass = 'DOIImportObject';
@@ -68,7 +69,7 @@ class DOIImporter extends IslandoraBatchImporter {
    *
    * @see IslandoraBatchImporter::readyForBatch()
    */
-  public static function readyForBatch(array & $form_state) {
+  public static function readyForBatch(array &$form_state) {
     $text = trim($form_state['values']['textarea']);
     $file = file_load($form_state['values']['file']);
     if ($file) {
@@ -95,7 +96,7 @@ class DOIImporter extends IslandoraBatchImporter {
    *
    * @see IslandoraBatchImporter::getBatchInfo()
    */
-  public static function getBatchInfo(array & $form_state) {
+  public static function getBatchInfo(array &$form_state) {
     $ids = $form_state['values']['ids'];
     return $ids;
   }
@@ -129,8 +130,7 @@ class DOIImportObject extends IslandoraImportObject {
     $headers = array_change_key_case($response->headers);
     if ($response->data &&
       strpos($headers['content-type'], 'text/html') === FALSE &&
-      strpos($response->data, "Malformed DOI") === FALSE
-    ) {
+      strpos($response->data, "Malformed DOI") === FALSE) {
 
       $crossref_xml = simplexml_load_string($response->data);
       // Totally don't support books so let's not attempt to make records that
@@ -168,4 +168,3 @@ class DOIImportObject extends IslandoraImportObject {
     return $this->mods;
   }
 }
-
--- a/modules/doi/modules/doi_importer/includes/importer.inc
+++ b/modules/doi/modules/doi_importer/includes/importer.inc
@@ -4,7 +4,6 @@
  * @file
  * Defines a class for importing DOI citations.
  */
-
 class DOIImporter extends IslandoraBatchImporter {
 
   protected $itemClass = 'DOIImportObject';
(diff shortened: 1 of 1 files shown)


bryjbrown:
@DiegoPino Not sure what the Travis failures are, I don't see any specific errors in the logs pointing me towards a mistake a I can quickly rectify. Advice?

bryjbrown:
@DiegoPino nvm I think I see the issue. I ran `drush coder-format` on the file before committing and it looks like it made a bunch of stupid changes (like s/&$var/& $var/). Reverting them now.

DiegoPino:
@bryjbrown ok, i see  this (twice)
```Shell
sites/all/modules/islandora_scholar/modules/doi/modules/doi_importer/includes/importer.inc:
 +6: [minor] File doc comments must be followed by a blank line.
 +29: [normal] missing space after comma
 +29: [minor] Comma not allowed after last value in single-line array declaration
 +29: [minor] Last item of an inline array must not be followed by a comma
 +172: [minor] Files must end in a single new line character
````
Not sure why `dcs` are running twice for each file... maybe it's just this Travis run. Will have to check closer, could be all just ducktaped together ?

**Comment by bryjbrown** [comment]
Thats what I get for copy-pasting from Stack Overflow and trusting in `drush coder-format` :)

**Comment by DiegoPino** [comment]
@bryjbrown seems like Travis for this whole module is a bit redundant. Its all about the submodules, not sure if drush coder even cares about that (i mean excludes submodules when running on a parent one) 

Look: first run -> https://travis-ci.org/Islandora/islandora_scholar/jobs/205837752#L905

Then this one https://travis-ci.org/Islandora/islandora_scholar/jobs/205837752#L946 runs on the same errors again.

**Comment by bryjbrown** [comment]
@DiegoPino Yep, it looks like a `drush coder-review` of islandora_scholar runs recursively on all the submodules, so running `drush coder-review` on each submodule is redundant as you say.

`drush coder-review` is being run on each submodule from [islandora_scholar/.travis.yml](https://github.com/Islandora/islandora_scholar/blob/7.x/.travis.yml#L48-L59), should these extra lines be deleted?

On the one hand redundancy isn't hurting anything, but Travis checks may run faster w/o them.

**Comment by DiegoPino** [comment]
@bryjbrown, thanks. Lets leave that redundancy question to this week's call then. Having a faster Travis is good especially for releases! But really does not hurt to check and recheck.

File: modules/doi/modules/doi_importer/includes/importer.inc [hunk 1]
```diff
@@ -24,38 +24,44 @@ class DOIImporter extends IslandoraBatchImporter {
...
-        '#title' => t('DOI Batch Journal Importer'),
-        '#description' => t('Provide either a list of Digital Object Identifiers in a file, or enter DOIs in the textarea below. In either case, the DOIs in the list should be separated by either whitespace or commas (or some combination thereof). NOTE: Presently, only journal (articles) are supported.'),
-        'file' => array(
-          '#type' => 'managed_file',
-          '#title' => t('File of DOIs'),
-          '#description' => t('A file listing Digital Object Identifiers (with ".txt" extension).'),
-          '#upload_validators' => array(
-            'file_validate_extensions' => array('txt'),
+  public static function getForm(array & $form_state) {
+    $openurl_pid = variable_get('islandora_doi_openurl_pid', 'user@example.com');
```
DiegoPino, line 28: I just hope nobody in islandora uses `example.com` email addresses !! 😄

DiegoPino: Just to be sure  "OpenURL PID email address" == "crossref registered email address". We do have somewhere that explains this or is this a thing Scholar module users are 100% aware?

DiegoPino (approved):

modules/doi/modules/doi_importer/includes/importer.inc
bryjbrown: @DiegoPino It is NOT something Scholar users are aware of, since both of the current module maintainers (@DonRichards  and I) encountered the same issue where we expected the DOI Importer to work out of the box and didn't know you had to register an email address and add it in the DOI config admin menu.

This is the primary reason why I chose to have the form not load if the email is set to the default; if the email is the default, they probably don't know that they even need to change it. The error message that gets displayed has a link to the page where they can set that configuration, and that page has more info how to set up a CrossRef registered email account.

Theres the chance that they could be using an email address other than the default which is not a registered CrossRef email, but we really have no way of checking that and if they enter a custom invalid email address than its their own fault ;)

DiegoPino: @bryjbrown I get it. We need more documentation for this probably. Will be good to get that sorted after code freeze. I wonder if we could do some test based on 
https://github.com/Islandora/islandora_scholar/blob/7.x/modules/doi/includes/utilities.inc#L38-L50
Basically, something that allows us to be sure that the stored email will work, and if so, let the whole thing pass. Just because, without knowing I could be tempted to put any email (not registered) to test wich would end triggering again a silent response here...

For now i will say ok, but we should add a note to the jira ticket  and open a new one that solves this for good. DOI-s are quite popular!

DiegoPino:
@bryjbrown @DonRichards will merge this. I'm quite sure this solves the issue with the default email, but not the one with any invalid email around. So opened this ticket to follow up on that. 
https://jira.duraspace.org/browse/ISLANDORA-1923

PR #259, opened 2017-03-16
Try unconditionally including namespace

jonathangreen:
Test to see if we can fix travis CI like this. **Shouldn't be merged.**


jonathangreen:
Oh yah. I'm just seeing if its possible now. Not recommending we do it :)

PR #260, opened 2017-03-30
ISLANDORA-1013 Allow Admin to toggle default embargo notifications

joshdentremont:
**JIRA Ticket**: https://jira.duraspace.org/browse/ISLANDORA-1013

Adds a checkbox to the admin form for the embargo module (admin/islandora/solution_pack_config/embargo) which allows Admin to turn off the default embargo emails. This allows Admin to create their own embargo notifications by using the Rules module, or implementing hook_islandora_scholar_embargo_users_to_notify() themselves.

# What's new?

Just the new variable which edits the behavior in islandora_scholar_embargo_islandora_scholar_embargo_users_to_notify(AbstractObject $object)

# How should this be tested?

Visit the admin form and try lifting embargoes with the checkbox toggled both to true and to false.

# Additional Notes:

We may wish to add to the documentation to explain how to create your own notifications if turning the defaults off.

# Interested parties
@Islandora/7-x-1-x-committers

### DIFF
--- a/modules/islandora_scholar_embargo/includes/admin.form.inc
+++ b/modules/islandora_scholar_embargo/includes/admin.form.inc
@@ -37,6 +37,12 @@ function islandora_embargo_admin(array $form, array &$form_state) {
     'title' => array('data' => t('Content Model')),
   );
   $form = array();
+  $form['email'] = array(
+    '#type' => 'checkbox',
+    '#title' => 'Use default email notifications for embargoes',
+    '#description' => 'Uncheck if you would like to manually set up notifications using rules, or by implementing hook_islandora_scholar_embargo_users_to_notify()',
+    '#default_value' => variable_get('islandora_scholar_embargo_default_emails', TRUE),
+  );
   $form['intro'] = array(
     '#type' => 'item',
     '#title' => 'Select Content Models for objects to be embargoed',
@@ -77,10 +83,12 @@ function islandora_embargo_admin_submit(array $form, array &$form_state) {
   switch ($op) {
     case 'Reset to default':
       variable_del('islandora_embargo_content_models');
+      variable_del('islandora_scholar_embargo_default_emails');
       break;
 
     default:
       $enabled = array_filter($form_state['values']['the_table']);
       variable_set('islandora_embargo_content_models', $enabled);
+      variable_set('islandora_scholar_embargo_default_emails', $form_state['values']['email']);
   }
 }
--- a/modules/islandora_scholar_embargo/includes/admin.form.inc
+++ b/modules/islandora_scholar_embargo/includes/admin.form.inc
@@ -39,9 +39,9 @@ function islandora_embargo_admin(array $form, array &$form_state) {
   $form = array();
   $form['email'] = array(
     '#type' => 'checkbox',
-    '#title' => 'Use default email notifications for embargoes',
-    '#description' => 'Uncheck if you would like to manually set up notifications using rules, or by implementing hook_islandora_scholar_embargo_users_to_notify()',
-    '#default_value' => variable_get('islandora_scholar_embargo_default_emails', TRUE),
+    '#title' => 'Suppress default email notifications for embargoes',
+    '#description' => 'Check if you have custom embargo notifications and would like to stop the default email notifications from sending.',
+    '#default_value' => variable_get('islandora_scholar_embargo_suppress_default_emails', FALSE),
   );
   $form['intro'] = array(
     '#type' => 'item',
@@ -83,12 +83,12 @@ function islandora_embargo_admin_submit(array $form, array &$form_state) {
   switch ($op) {
     case 'Reset to default':
       variable_del('islandora_embargo_content_models');
-      variable_del('islandora_scholar_embargo_default_emails');
+      variable_del('islandora_scholar_embargo_suppress_default_emails');
       break;
 
     default:
       $enabled = array_filter($form_state['values']['the_table']);
       variable_set('islandora_embargo_content_models', $enabled);
-      variable_set('islandora_scholar_embargo_default_emails', $form_state['values']['email']);
+      variable_set('islandora_scholar_embargo_suppress_default_emails', $form_state['values']['email']);
   }
 }
--- a/modules/islandora_scholar_embargo/islandora_scholar_embargo.install
+++ b/modules/islandora_scholar_embargo/islandora_scholar_embargo.install
@@ -12,7 +12,7 @@ function islandora_scholar_embargo_uninstall() {
   $variables = array(
     'islandora_embargo_content_models',
     'islandora_scholar_embargo_whitelisted_roles',
-    'islandora_scholar_embargo_default_emails',
+    'islandora_scholar_embargo_suppress_default_emails',
   );
   array_walk($variables, 'variable_del');
 }
--- a/modules/islandora_scholar_embargo/islandora_scholar_embargo.install
+++ b/modules/islandora_scholar_embargo/islandora_scholar_embargo.install
@@ -12,6 +12,7 @@ function islandora_scholar_embargo_uninstall() {
   $variables = array(
     'islandora_embargo_content_models',
     'islandora_scholar_embargo_whitelisted_roles',
+    'islandora_scholar_embargo_default_emails',
   );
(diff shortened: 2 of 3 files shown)


DiegoPino:
@J77D is this ready to be reviewed? will re-run travis because it was failing.

joshdentremont:
Thanks Diego. There is still one failing test. It doesn't look like the issue is on my end, but please let me know if it is.

manez:
Restarting that test to see if it was just a hiccup.

DiegoPino:
Hi @J77D catching up on this. Wonder what the reason is making the email notification check inside the hook instead in the hook(s) invocation? I wonder if moving that logic to the outside would allow other hooks to have consistently that option being enforced once set?

joshdentremont:
Hi @DiegoPino Thanks for the feedback. Do you mean to put the if statement around this line https://github.com/Islandora/islandora_scholar/blob/7.x/modules/islandora_scholar_embargo/islandora_scholar_embargo.module#L353 

I think we need the hook to at least return an empty array so that the drupal_alter on the next line works. If someone wanted to populate the users in code themselves they could use https://github.com/Islandora/islandora_scholar/blob/7.x/modules/islandora_scholar_embargo/islandora_scholar_embargo.api.php#L36 I believe. This way any existing hooks should still work.

Thanks

bryjbrown:
@J77D @DiegoPino Sorry for sitting on this one so long, I was waiting until after the 1.9 release to start merging in new code. Ready to review/test this now.

@J77D I get the WHAT of this PR from the message, but not the WHY. Is this meant to suppress the email to the object's owner (default behavior) so that if you use an alternate notification system you don't get double emails? Just trying to understand the context that made you want to write this. LMK the deal with this and I'll be happy to review/test ASAP! :+1:

DiegoPino:
@bryjbrown 
>I get the WHAT of this PR from the message, but not the WHY.

True. The "why" is a bit confusing, but for what I understand it seems like if you need a different way of informing about lifted embargos to users then having a hook + disabling internal notification could be desired? The change I asked for was to avoid this being run in the wrong spot, mostly trying to move check/if from the particular hook implementation to outside the hook invocation iteration to basically shut all emails if the toggle was disabled, which does not seem to be addressed ( 👀 @J77D) at https://github.com/Islandora/islandora_scholar/blob/7.x/modules/islandora_scholar_embargo/islandora_scholar_embargo.module#L353 (or near there) but maybe I'm not getting the logic correct 😁 

@bryjbrown will wait for your thoughts on this.

Best!

bryjbrown:
@DiegoPino @J77D I *think* that if you were to wrap https://github.com/Islandora/islandora_scholar/blob/7.x/modules/islandora_scholar_embargo/islandora_scholar_embargo.module#L353 in an if statement, it would stop ALL implementations of `hook_islandora_scholar_embargo_users_to_notify()` from being called, not just the default one. `module_invoke_all()` [calls all enabled modules that make use of the defined hook](https://api.drupal.org/api/drupal/includes%21module.inc/function/module_invoke_all/7.x), so the `$users` coming out of this represent the users returned by all the separate implementations of `hook_islandora_scholar_embargo_users_to_notify()`. IMHO having the check in the default hook is the right way to go, but I'm not confidant about it and could use a third opinion. 

@dannylamb or any other @Islandora/7-x-1-x-committers, could you take a look at this and give an opinion?

DiegoPino:
@bryjbrown true, maybe i got the use case wrong (and we need to update that jira ticket also, points to another need really). I thought the idea was to have a silencing toogle, so any notifications are indeed silenced and not only this particular hook. But now that i read it from your words, your logic makes fully sense to me 😄 thanks!

DonRichards:
I could totally use this in our IR. We're developing an unmediated IR and the final step would be nice to edit or bypass.

joshdentremont:
@bryjbrown The WHY is as you say, to disable the extra emails. These default emails go to the Drupal account with the ID of 1, and the person that uploaded the object being embargoed. In our case these are usually both the wrong person to send embargo emails to, so we would prefer to have those emails not be sent.

I prefer to use Rules instead of using the hook. Adding this option to the admin form lets me disable the default hook, set up whatever rules I like, and not worry about other people getting emails.

@DiegoPino had the suggestion to put the if statement around https://github.com/Islandora/islandora_scholar/blob/7.x/modules/islandora_scholar_embargo/islandora_scholar_embargo.module#L353 which would stop the hook from being invoked at all. This would also work for us, but my thought was that doing this would break any custom hook implementations or alter implementations that people currently have set up, if they toggled the defaults off. 

Thinking about this now, anyone who has a custom hook implementation would just have to leave the defaults enabled. I'm OK with putting the if statement here if that is what people prefer. We would want to be clear in the admin form that if someone toggles this they would lose any email notifications that they have custom hooks for but I don't see that being an issue. I can change the title and description of the option to reflect this change if needed.

Thanks for the discussion & feedback

bryjbrown:
@J77D My vote would be to keep your check inside the default hook. I have future plans to use this hook in other modules, and being able to turn off the default hook while still allowing external ones to run would be very useful. Now that I understand the intent of this PR I'm on board.

I'll try to give this a test sometime this week.

bryjbrown:
@J77D @DiegoPino As promised, I finally found some time this afternoon and gave this PR a test, and it works as expected. By default emails get sent out but if you turn off the default notifications emails DO NOT get sent out.

I even went a bit further and tested using the shell of an embargo notification manager module I've been cobbling together, and sure enough turning off default notifications does NOT stop external implementations of hook_islandora_scholar_embargo_users_to_notify() from running.

But in giving this an actual human run through, I did see something that I'd like to see improved before I can approve: the language used to describe this feature. I feel that its a bit inaccurate and confusing.
```
Use default email notifications for embargoes
Uncheck if you would like to manually set up notifications using rules, or by implementing hook_islandora_scholar_embargo_users_to_notify()
```
This wording makes it sound like you have to uncheck the box if you want to use notification methods other than the default, when in reality you can aggregate a number of methods for building the list of users to notify. Some people may want to use their own custom hook AND the default,

I would at least like to see the help text updated a bit to say that unchecking this option simply suppresses the default notification, because the current wording makes it sound like you MUST uncheck this option in order to use other notifications. 

IDEALLY I think the best and most clear version of this option would be to switch the Boolean (although this would take a bit more work to flip the bool checking in the code and require more testing) so that the option says something like 
```
Suppress default email notifications for embargoes
Check if you have custom embargo notifications and would like to stop the default email notifications from sending.
```
and would be unchecked (FALSE) by default. It would be much more clear to users what they are doing when they check the box than the current version IMHO, although I'm open to discussion on this and perhaps asking a few scholar embargo users to weigh in.

joshdentremont:
Thanks @bryjbrown I agree that the wording you suggest would be better. Did you want to discuss it with others before I make the change?

bryjbrown:
@J77D not a bad idea. @DiegoPino @DonRichards do you have any opinions or suggestions on the wording of this config option? See https://github.com/Islandora/islandora_scholar/pull/260#issuecomment-304383515 for context.

DiegoPino:
@J77D I agree with @bryjbrown about the boolean change and also the wording change need. Sadly wording is not my stuff, can hardly make others understand what I'm trying to say 😬, so would ask any other one to suggest something there.

DonRichards:
```shell
Suppress default email notifications for embargoes
```
@bryjbrown @DiegoPino @J77D This is a win. It's clear and I think the other way would be confusing. Using "Suppress" really hits home what the function of this is. Thumbs up from me.

DiegoPino:
@J77D @DonRichards @bryjbrown seems to be agreement on the wording then 😄 
Once that hits as a commit (and Travis-CI passes) i would say you are all set to merge (you should review and approve first of course!)

bryjbrown:
@J77D Awesome! An initial glance at the new code looks like you nailed it, but I'd still like to run the tests again to make sure the logic checks out. I'll try to get that taken care of ASAP.

joshdentremont:
Awesome. Thanks for your help on this @bryjbrown and for the feedback Don and Diego

bryjbrown:
@J77D Just tested this with the same series of actions I did before the variable's value was reversed, and everything is :+1:. Merging.

bryjbrown (approved):

PR #261, opened 2017-04-06
Update phpunit test cases to use new namespaced functions.

jonathangreen:
## JIRA Ticket
https://jira.duraspace.org/browse/ISLANDORA-1935

## What does this Pull Request do?

Update to use namespaces in PHP unit so tests with PHP7 don't fail.

Some good info here:
https://thephp.cc/news/2017/02/migrating-to-phpunit-6

> While PHPUnit 5.7 has a forward compatibility layer, meaning you can already use PHPUnit\Framework\TestCase with that version, there is no backward compatibility layer in PHPUnit 6 that allows you to use

## How should this be tested?

Make sure travis doesn't fail, if it passes we should be all good.

## Interested parties
@DiegoPino

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,6 @@
 sudo: required
 language: php
 php:
-  - 5.3.3
-  - 5.4
-  - 5.5
   - 5.6
   - 7.0
   - 7.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 sudo: required
 language: php
 php:
+  - 5.3
+  - 5.4
+  - 5.5
   - 5.6
   - 7.0
   - 7.1
@@ -38,6 +41,8 @@ before_install:
 before_script:
   # Mysql might time out for long tests, increase the wait timeout.
   - mysql -e 'SET @@GLOBAL.wait_timeout=1200'
+  - composer require phpunit/phpunit '>=4.8.35|>=5.4.3.,<=5.0'
+  - composer install
 script:
   - ant -buildfile sites/all/modules/islandora_scholar/build.xml lint
   - sites/all/modules/islandora_scholar/tests/scripts/line_endings.sh sites/all/modules/islandora_scholar
@@ -62,4 +67,4 @@ script:
   - mv $HOME/tcpdf $TRAVIS_BUILD_DIR/modules/exporter/lib/
   - cp -Rv $JAVASCRIPT_DIR/* $TRAVIS_BUILD_DIR
   - phpcpd --names *.module,*.inc,*.test sites/all/modules/islandora_scholar
-  - phpunit sites/all/modules/islandora_scholar/modules/citeproc/tests/CSL_Dateparser.test
+  - vendor/bin/phpunit sites/all/modules/islandora_scholar/modules/citeproc/tests/CSL_Dateparser.test
--- a/modules/citeproc/tests/CSL_Dateparser.test
+++ b/modules/citeproc/tests/CSL_Dateparser.test
@@ -6,12 +6,14 @@
 
 require_once dirname(__FILE__) . '/../includes/CSLDateparser.php';
 
+use \PHPUnit\Framework\TestCase;
+
 /**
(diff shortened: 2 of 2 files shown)


jonathangreen:
Hrm. Soo I guess this isn't as smooth as I thought. We can support all currently supported PHP versions like this.
http://php.net/supported-versions.php 

5.6
7.0
7.1

But we can't support < 5.6

jonathangreen:
@Islandora/7-x-1-x-committers what does everyone think of this? It will get our tests passing again, but we are not testing older versions of PHP anymore. Should probably be discussed at next committers call. 

Same deal here:
https://github.com/Islandora/tuque/pull/155

manez:
I like the idea of reducing our obligations to older versions of PHP, but this is indeed a big enough deal that we should spend more time and discussion. Maybe a thread on the listserv to catch the attention of people who might be relying on older PHP version and not following Committers' Calls?

DiegoPino:
I agree with @manez , list and again some committers talk. I tried to propose such an "crazy" idea (deprecating 5.3) during the previous release process and some of the reasons i got for not doing it were understandable. I agree that having PHP7 code tested in year 2017 is imperative, but Drupal 7 really has never been officially supported (as fas as i know) under that version, which leads to my following conclusion, deprecating 5.3 is fine because nothing runs really anymore there, but 5.5 not sure because, well Drupal supports it. I'm pretty sure we can alternatives ways for testing (2 classes) or even, not testing that class at all for 7. (or for 5.5 and 5.3)

pc37utn:
RHEL 6 is supporting PHP 5.3 ( althought with backports from newer versions) until 2020 and RHEL 7 will be with PHP 5,4 until 2024.  Not that we aren't testing a remi repo to add PHP 7 now, but Red Hat/CentOS are using/supporting the older versions.

jonathangreen:
> I agree that having PHP7 code tested in year 2017 is imperative, but Drupal 7 really has never been officially supported (as fas as i know) under that version

D7 does officially support PHP 7 
https://www.drupal.org/docs/7/system-requirements/php#php_required

In the Drupal 7.50 release notes it is mentioned:
https://www.drupal.org/blog/drupal-7-50

And people are even encouraged to update, so I would say D7 is on the PHP7 train.

I will say that I as against dropping old versions of PHP because of RHEL support before, but Zend officially saying older versions of PHP are unsupported made me change my mind. I just thought I'd propose a fix here for our testing issues. Definitely needs some wider discussion.

jonathangreen:
@manez I made a post to the list to get some more discussion.

DiegoPino:
@jonathangreen Cool. I had not read that doc and don't get me wrong. I support and use PHP7. But still, php 5.5 is also still out there and supported. Not sure if the discussion here is if PHP7 is or not something we need, we need it and we have supported lately all pulls that sought to make islandora compat. But if we are giving away 5.5 for the sake of php 7, we need to have consensus. Also, my opinion, the Drupal statement is kinda liberal. Maybe core supports it, but popular modules Like Views would not even pass a simple Drupal Coding Standards check with those Class naming conventions.

jonathangreen:
@DiegoPino thats fair and I agree we need consensus. Honestly, I somewhat regret opening this pull request. I was just trying to fix the issues with tests, and I thought I could do it it for all versions of PHP and it turned into only working with PHP 5.6+ so I posed this question. I wasn't really intending to be the poster boy for removing testing of old versions of PHP. 😄

DiegoPino:
@jonathangreen no regrets. I feel this pull request pushes the discussion to a resolution. Grateful for that, community needs some "incentive" sometimes to advance. I'm pretty sure we will come a solution that makes all happy and allows us, once for all, fix and move forward. Thanks a lot.

jordandukart:
PHP5.3+ added back in again.

DiegoPino:
@jonathangreen @jordandukart will merge this once Travis is calm (give me an hour or so). Lots of readme related pulls being merged right now,

DiegoPino (approved):

PR #263, opened 2017-04-19
Update README.md - fix typo

manez:
# What does this Pull Request do?

Fixes a typo.

# What's new?
_moudle_ becomes _module_.

# How should this be tested?

Review the change. Confirm that 'moudle' is not a thing but 'module' is.

# Interested parties
@bryjbrown or @DonRichards want to give it a go?

PR #264, opened 2017-04-19
Update README.md - fix typo (release version)

manez:

Fixes a typo. Release version of #263 

# What's new?
_moudle_ becomes _module_.

# How should this be tested?

Review the change. Confirm that 'moudle' is not a thing in 7.x-1.9 either. 

# Interested parties
@bryjbrown or @DonRichards again

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ This module requires the following modules/libraries:
 * [Citeproc](https://github.com/Islandora/islandora_scholar/tree/7.x/modules/citeproc) (included in /modules)
 * [CSL](https://github.com/Islandora/islandora_scholar/tree/7.x/modules/csl) (included in /modules)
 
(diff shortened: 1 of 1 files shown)


bryjbrown:
@DiegoPino @manez How come this release pull doesn't require a review? I would approve it anyway, just think its strange since usually its there.

manez:
@bryjbrown we don't have that feature turned on in the release branch

ETA: Or maybe we do, just not consistently?

DiegoPino:
Makes pushing (force!) easier for the release manager.
-- 
Diego Pino Navarro
Digital Repositories Developer
Metropolitan New York Library Council (METRO)

DonRichards:
The Travis failure is unrelated. It's obviously not failing the README file. I'll merge. But I do think if it hasn't already been added the "Review Requirement" needs to be set. I don't have access to the settings tab for this.

DiegoPino:
@DonRichards: I have many reasons for not making that more complicated than needed but I understand your concerns.
During a release process and after code freeze, all pull requests that aspire to be part of the release, are done in pairs: which means that if the parent one, the one that goes into 7.x was reviewed and was accepted, the release branch one, which is identical and should only be merged if it is actually identical should not need review again, making this process a bit more agile.
But the most important is to be able, as release manager, to change version tags (inside .info files) and even push force the whole release branch. If Review Requirement is present, I think (and I can be wrong) that seems more complicated. 
But nothing is lost. I can enable review if you feel it is very needed but would suggest to propose this to all committers during a call.

PR #278, opened 2018-03-02
Adjusts menu labels/descriptions

bryjbrown:
**JIRA Ticket**: https://jira.duraspace.org/browse/ISLANDORA-1284

Renames "Islandora Scholar Embargoes" to "Scholar Embargoes" in the Islandora Solution Pack menu so that the two appear together, instead of "Islandora Scholar Embargoes" appearing way up the list before "Scholar".

An addition part of this ticket involves a request to put a copy of the "Manage Embargoed Items" inside the Scholar admin menu, but I don't agree with this and didn't implement it. Adding menu tabs for one module into a separate module's admin screen seems messy.

# How should this be tested?
Boot up a VM and go to /admin/islandora/solution_pack_config/. You'll see that "Scholar" and "Scholar Embargoes" now appear next to each other. 

Note: If you booted up with the original Scholar and then switch to this PR, you'll need to clear the cache to rebuild the menu cache.

# Interested parties
@DonRichards @DiegoPino @dmoses @bondjimbond @rosiel

### DIFF
--- a/islandora_scholar.module
+++ b/islandora_scholar.module
@@ -12,7 +12,7 @@ function islandora_scholar_menu() {
   return array(
     'admin/islandora/solution_pack_config/scholar' => array(
       'title' => 'Scholar',
-      'description' => 'Configure scholar settings such as RoMEO and Google Search',
+      'description' => 'Configure Islandora Scholar settings such as RoMEO and Google Search',
       'type' => MENU_NORMAL_ITEM,
       'page callback' => 'drupal_get_form',
       'page arguments' => array('islandora_scholar_admin_form'),
--- a/modules/islandora_scholar_embargo/islandora_scholar_embargo.module
+++ b/modules/islandora_scholar_embargo/islandora_scholar_embargo.module
@@ -26,8 +26,8 @@ function islandora_scholar_embargo_menu() {
     'access arguments' => array(2),
     'file' => 'includes/embargo.inc',
   );
-  $items['admin/islandora/solution_pack_config/embargo'] = array(
-    'title' => 'Scholar Embargo',
+  $items['admin/islandora/tools/embargo'] = array(
+    'title' => 'Islandora Scholar Embargo',
     'description' => 'Configure the Islandora Scholar Embargo module.',
     'page callback' => 'drupal_get_form',
     'page arguments' => array('islandora_embargo_admin'),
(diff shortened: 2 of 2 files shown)


bryjbrown:
Pinging @Islandora/7-x-1-x-committers

bondjimbond:
Possibly out of scope for this, but I'd be tempted to change that name to "Islandora Embargoes" and cut out the Scholar connection altogether, given that you can embargo anything (not just Citation and Thesis objects). The name "Scholar embargoes" confuses even some pretty advanced Islandora devs as to what it can be used for.

bryjbrown:
@bondjimbond I :100:% agree with you in theory, but in practice `islandora_scholar_embargo` is the module's name, so we'd have to change every hook function name in the module too. Or, having a module called "Islandora Embargo" with a bunch of hooks that say `islandora_scholar_embargo`. Either situation isn't ideal. There's also the fact that changing the name might introduce even MORE confusion to the community. If the word "Scholar" in the name has any utility, its in differentiating this module from "Islandora IP Embargo". 

I think the lesson to be taken away from all this is that naming is hard but very important, and lets make sure we give things good names in CLAW :crying_cat_face:

mjordan:
> I think the lesson to be taken away from all this is that naming is hard but very important, and lets make sure we give things good names in CLAW

+1

rosiel (commented):
Since it relates to a feature that could apply to any content type, and not to a specific content type (or two in the case of Scholar), would it make more sense for Embargoed objects to be under Islandora Utility Modules? (Not a blocker and feel free to disregard).

DiegoPino:
@bryjbrown that is a good question  @rosiel brings here. Should we group - and label - functionality by which module provides it (dev-inclined) or by its global islandora logic (user-inclined)?

bryjbrown:
@rosiel I would agree with that, having it appear in the SP menu always seemed odd to me. Islandora IP Embargo is in the Utility menu, and so is CSL Styles, a submodule of Scholar. I wouldn't be opposed to making such a change as part of this PR, but the downside of this is messing with established behavior (eg, people not being able to find the Scholar Embargo admin pages under the SP menu and thinking something is broken).

DiegoPino:
@bryjbrown if you feel it is time (you know this will eventually happen but I won't push it right now) to move those menus to their rightful place you got my support. I think most of the downsides/issues can be fixed via good documentation and release notes. Its up to you 👍 ⭐️

bondjimbond:
++ for moving the menu! It would help at least a bit with the confusion over how the tool can be used.

bryjbrown:
Alrighty, the Scholar Embargo admin page has been moved to the "Islandora Utility Modules" menu.

DiegoPino (approved):
Looks good to me. 24 hours so any other @Islandora/7-x-1-x-committers can comment on this change. Thanks @bryjbrown

PR #282, opened 2018-03-19
Updates thesis forms: s/Degree grantor/Degree granting institution/

bryjbrown:
**JIRA Ticket**: https://jira.duraspace.org/browse/ISLANDORA-1988

Changes the `thesis_form.xml` and `thesis_form_ISLANDORA1350.xml` thesis forms to say "Degree granting institution" instead of "Degree grantor" in the form label and output value. Also enhances the form's xpath search so that MODS records generated with "Degree grantor" as the `roleTerm` still get found.

Note that only newly created 'Degree granting institution' names will have the newer `roleTerm`, older ones will still get loaded by the form, but will retain their original 'Degree grantor' `roleTerm`. I'm open to suggestions on ways to force the form to update this data though.

# How should this be tested?
Load up a VM and create a thesis object using the default thesis form. Make sure to add a "Degree granter", and inspect the produced MODS record. Then, load up this PR locally and edit the MODS record with the default form. You will see the original data you entered, but the label will be "Degree granting institution" instead. Create a new DGI, and once again inspect the created MODS record. The new DGI should have the new proper `roleTerm`, while the older DGI has the "Degree grantor" `roleTerm`.

# Additional Notes:
If anyone knows how to get the form to update pre-existing info and resave it with the new MODS attributes, LMK and I'll work this into the PR.

# Interested parties
@DiegoPino @bondjimbond @rosiel @DonRichards

### DIFF
--- a/xml/thesis_form.xml
+++ b/xml/thesis_form.xml
@@ -478,7 +478,7 @@
           <multiple>FALSE</multiple>
           <required>FALSE</required>
           <resizable>FALSE</resizable>
-          <title>Degree Granter</title>
+          <title>Degree Granting Institution</title>
           <tree>TRUE</tree>
         </properties>
         <children>
@@ -501,10 +501,10 @@
                   <schema/>
                   <type>xml</type>
                   <prefix>NULL</prefix>
-                  <value>&lt;name type='corporate' authority='lcnaf'&gt;&lt;role&gt;&lt;roleTerm authority="marcrelator" type="text"&gt;Degree grantor&lt;/roleTerm&gt;&lt;/role&gt;&lt;/name&gt;</value>
+                  <value>&lt;name type='corporate' authority='lcnaf'&gt;&lt;role&gt;&lt;roleTerm authority="marcrelator" type="text"&gt;Degree granting institution&lt;/roleTerm&gt;&lt;/role&gt;&lt;/name&gt;</value>
                 </create>
                 <read>
-                  <path>mods:name[@type='corporate' and @authority='lcnaf']/mods:role[mods:roleTerm = 'Degree grantor']/..</path>
+                  <path>mods:name[@type='corporate' and @authority='lcnaf']/mods:role[mods:roleTerm = 'Degree grantor']/.. |  mods:name[@type='corporate' and @authority='lcnaf']/mods:role[mods:roleTerm = 'Degree granting institution']/..</path>
                   <context>parent</context>
                 </read>
                 <update>NULL</update>
--- a/xml/thesis_form.xml
+++ b/xml/thesis_form.xml
@@ -501,13 +501,16 @@
                   <schema/>
                   <type>xml</type>
                   <prefix>NULL</prefix>
-                  <value>&lt;name type='corporate' authority='lcnaf'&gt;&lt;role&gt;&lt;roleTerm authority="marcrelator" type="text"&gt;Degree granting institution&lt;/roleTerm&gt;&lt;/role&gt;&lt;/name&gt;</value>
+                  <value>&lt;name type='corporate' authority='lcnaf'&gt;%value%&lt;/name&gt;</value>
                 </create>
                 <read>
                   <path>mods:name[@type='corporate' and @authority='lcnaf']/mods:role[mods:roleTerm = 'Degree grantor']/.. |  mods:name[@type='corporate' and @authority='lcnaf']/mods:role[mods:roleTerm = 'Degree granting institution']/..</path>
                   <context>parent</context>
                 </read>
-                <update>NULL</update>
+                <update>
+                  <path>self::node()</path>
+                  <context>self</context>
+                </update>
                 <delete>
                   <path>self::node()</path>
                   <context>self</context>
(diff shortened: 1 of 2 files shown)


bondjimbond:
Thanks, @bryjbrown! This has been a long time coming! I'll give it a test and report back; code looks just fine, and I don't anticipate any issues.

bondjimbond:
Test results:

Create a new Thesis with this version: No problem; roleTerm is "Degree granting institution" as hoped for.

Create a new Thesis with old form, edit with this version: Problem.
- Form still (somehow?) reads the name from "Degree grantor", so the fields are populated as usual
- If you update, it still writes the updates to the element with the roleTerm "Degree grantor". "Degree granting institution" is not added anywhere to the MODS.

DiegoPino:
@bondjimbond i don't think that can be fixed or better said, should be a concern. Reading a newly well-formed XML doc created with this new form using the old form:  The old form simply won't work and I'm Ok with that. Is there a use case for that?

bryjbrown:
@bondjimbond The form populates the "Degree granting institution" field using both "Degree grantor" AND "Degree granting institution"; I thought this was desired behavior so that folks wouldn't lose access to existing data. The xpath that is used to find the data for this field is actually two xpaths connected with a `|` so that it grabs both. Unfortunately I can't figure out how to tell the XML form that after loading that data it should be flushed back into the MODS record using the new value. It seems to always save the data back to whereever it was loaded from if it existed before it was loaded into the form.

Any new "Degree granting institutions" created in the form, even if they exist alongside pre-existing "Degree grantor"s, will have the new correct `roleTerm`.

bryjbrown:
@bondjimbond @DiegoPino Here is the exact `<path>` the XML form is using to find and read data for populating the field. Notice the `|` character which joins the two separate xpaths into one, so that it will find the data whichever way it occurs as a courtesy to legacy records.

```
<path>mods:name[@type='corporate' and @authority='lcnaf']/mods:role[mods:roleTerm = 'Degree grantor']/.. |  mods:name[@type='corporate' and @authority='lcnaf']/mods:role[mods:roleTerm = 'Degree granting institution']/..</path>
```

bondjimbond:
@bryjbrown @DiegoPino What I was trying to say about my tests is that the new roleTerm "Degree Granting Institution" does not appear when you edit an existing Thesis object, only the original "Degree Grantor"... I think our goal was to have parallel ones, but what we end up with is no change to the old docs.

But perhaps I'm misreading the testing instructions? Is the goal here to create entries with two roleTerms in new objects, and leave old objects unchanged when you edit them?

bryjbrown:
@bondjimbond What I'm trying to say is that I can only find a way to give the new `roleTerm` to Degree Granting Institutions added AFTER you switch to the updated form. I cannot figure out how to "fix" the old ones. Any DGIs described prior to using this form will stay the same. I'd like to figure out a way to update them on save, but don't know how.

And I'm not sure if I agree about the original goal being to have parallel ones. Having information created by the form but that you no longer have access to through the form seems wrong to me, so I tried to retain access to the legacy `roleTerm` via the xpath.

bondjimbond:
@bryjbrown Just to make sure we understand each other, 'cause I think we're having a wee communication failure..

> What I'm trying to say is that I can only find a way to give the new roleTerm to Degree Granting Institutions added AFTER you switch to the updated form

Yes, that's what I'm getting at... After I switch to the new form and edit the object that had been originally ingested with the old form, it won't save the new roleTerm. 

>   I cannot figure out how to "fix" the old ones. Any DGIs described prior to using this form will stay the same. I'd like to figure out a way to update them on save, but don't know how.

Are we saying the same thing, then?

If the case is that you're trying to read from both "degree grantor" and "degree granting institution", but can't save to just "degree granting institution", perhaps it would be better to drop the attempt to read the old value using the new form? Someone editing the metadata would have to notice the DGI field is blank and fill it out. Then the values with Degree Grantor would remain in the MODS XML, but they wouldn't bother the new metadata...

It's not great, perhaps better than the form only working properly on new objects?

bryjbrown:
@bondjimbond Completely ignoring previously created Degree grantors in the records so that only newly created information is recognized is certainly one way to solve the problem, but I don't think its the best way. I think people should still be able to access and edit the legacy info via the form. There MUST be way to to make sure that whenever a pre-existing field is updated, it gets saved to the new location.

Reading through the XML forms documentation (https://github.com/Islandora/islandora/wiki/Creating-and-Working-With-XML-Forms), it looks like could be a way to do this via the `<update>` element, but I'm not sure how. I've tried a few different ways of configuring it but none of them seem to have any effect. @DiegoPino @rosiel do either of you know any XML forms gurus who could help us configure the form to save this data to the right place on update?

bondjimbond:
@bryjbrown Just made a clone of your version of the form so that I could view it in the Form Builder. The "Update" box isn't checked at all, so I expect this is why it isn't updating the changes. Maybe the "Update" should go straight to "degree granting institution" and ignore "degree grantor"?

bryjbrown:
@bondjimbond So my understanding is that its READING from both places, but the output value should just be "Degree granting institution". I never mess with XML forms, so I'll take a look at the form builder UI to see if I can work something out.

DiegoPino:
@bondjimbond for a reason i can't explain, if update is not present, it gets disabled on XML form builder interface (BUG). You need to add it via text editor and upload, then you get it appears again in XML form builder 😢

bryjbrown:
@bondjimbond Figuring out what to put in the update is kinda hard, its not just checking the update box but configuring how the info should get written back to the document, which is tough because of the way everything is nested. Working on it more now, hopefully we'll get some gurus to guide us.

bryjbrown:
So everyone should know that @rosiel is an XML forms genius from beyond this plane of existence. She whipped up a fix that will read both "Degree grantor" and "Degree granting institution", but write both as "Degree granting institution" back to the MODS. The form actually fixes the old metadata! This is the fix I was hoping for. I tested this myself, but @bondjimbond should test as well to make sure we're all on the same page.

Testing instructions:
Create an object while on the 7.x branch to use the old form. After you have created the object, verify that it says "Degree grantor" in the metadata. Then switch to this PR branch, and load the object's MODS into the editor and save. "Degree grantor" should be changed to "Degree granting institution" in the MODS.

:clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap:

bondjimbond:
Brilliant work, @rosiel and @bryjbrown! It works for me.

bondjimbond (approved):
👍

DiegoPino:
Cool. @bryjbrown while testing I saw that the <mods:originInfo>
 <mods:dateIssued keyDate="yes">, which is a required field, has this description: "The year this book was issued." Since you are already here, could you change that to something relevant to thesis?  Or is book the right one there? I don't deal too much with thesis really

bondjimbond:
@DiegoPino  I noticed that as well when working with the form... This was probably originally copied from another form where "Book" was more appropriate. It should be changed, but not in scope for this PR - so if @bryjbrown wants to tweak that wording, awesome; but otherwise this is good to go anyway.

bryjbrown:
@DiegoPino @bondjimbond :+1: Got it

DiegoPino:
>but otherwise this is good to go anyway.

@bondjimbond Of course! That is why I did not request a change. It was just a comment.

rosiel:
Lookin for a merge... @bondjimbond or any other @Islandora/7-x-1-x-committers

PR #284, opened 2018-03-23
Revert "Updates thesis forms: s/Degree grantor/Degree granting institution/"

adam-vessey:
Reverts Islandora/islandora_scholar#282

JIRA Ticket: https://jira.duraspace.org/browse/ISLANDORA-1988

The present approach taken in the pull request being reverted (updating metadata on the fly, only when editing with particular forms) is likely to lead to a significant period in which there is inconsistent metadata present in repositories. Additionally, there appear to be multiple uses of the deprecated role term "Degree grantor" elsewhere in the code base, in particular, two submodules of `islandora_scholar`:
* ["islandora_google_scholar" XPath looks for "Degree grantor" explicitly](https://github.com/Islandora/islandora_scholar/blob/6494ac2082a6fbf36e637ee716c0e21d107b87a6/modules/islandora_google_scholar/islandora_google_scholar.module#L176)
* ["citeproc" will map "dgg" as the old "Degree grantor" instead of the new "Degree granting institution"](https://github.com/Islandora/islandora_scholar/blob/b148d4a5f0a42cedfca87556fb0cb8f57499cf36/modules/citeproc/includes/marcrelator_conversion.inc#L82)

There's also some mention of this role term in [islandora_xml_forms](https://github.com/Islandora/islandora_xml_forms/blob/39674ce770bf52f803e5417f62d5e7682e7ec78b/elements/xml/relators.rdf#L1287)... should this be updated in some way? Is this file even used?

To minimize the window in which there can exist inconsistent metadata, it is likely preferable to update all entries with the deprecated value to the preferred in a batch process. This has the side effect of allowing the code where we _use_ the value to be simplified such that we only need to deal with the preferred value; though, it may be desirable to introduce configuration such that the old value may continue to be used by institutions which wish to delay updating their data. The batch process should probably follow much the same patterns as we have followed in other instances of updating data, with a [message emitted from a `hook_update_N()` implementation pointing at a drush script](https://github.com/Islandora/islandora_xacml_editor/blob/ed2cb7b5ad036470afa2e763d12c13849c611568/api/islandora_xacml_api.install#L396-L407).

In the interim, before this hypothetical batch process is written and ready, and other code updated to use the new value, it seems prudent to revert [the merged pull request](https://github.com/Islandora/islandora_scholar/pull/282).

Tagging, due to nature of issue and potential for inconsistencies popping up in the wild: @dannylamb

### DIFF
--- a/xml/thesis_form_ISLANDORA1350.xml
+++ b/xml/thesis_form_ISLANDORA1350.xml
@@ -443,7 +443,7 @@
           <multiple>FALSE</multiple>
           <required>FALSE</required>
           <resizable>FALSE</resizable>
-          <title>Degree Granting Institution</title>
+          <title>Degree Granter</title>
           <tree>TRUE</tree>
         </properties>
         <children>
@@ -466,16 +466,13 @@
                   <schema/>
                   <type>xml</type>
                   <prefix>NULL</prefix>
-                  <value>&lt;name type='corporate' authority='lcnaf'&gt;%value%&lt;/name&gt;</value>
+                  <value>&lt;name type='corporate' authority='lcnaf'&gt;&lt;role&gt;&lt;roleTerm authority="marcrelator" type="text"&gt;Degree grantor&lt;/roleTerm&gt;&lt;/role&gt;&lt;/name&gt;</value>
                 </create>
                 <read>
-                  <path>mods:name[@type='corporate' and @authority='lcnaf']/mods:role[mods:roleTerm = 'Degree grantor']/.. |  mods:name[@type='corporate' and @authority='lcnaf']/mods:role[mods:roleTerm = 'Degree granting institution']/..</path>
+                  <path>mods:name[@type='corporate' and @authority='lcnaf']/mods:role[mods:roleTerm = 'Degree grantor']/..</path>
                   <context>parent</context>
                 </read>
-                <update>
-                  <path>self::node()</path>
-                  <context>self</context>
-                </update>
+                <update>NULL</update>
                 <delete>
                   <path>self::node()</path>
                   <context>self</context>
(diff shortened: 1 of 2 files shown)


bryjbrown:
@adam-vessey Good call, I wasn't aware of these other places where Degree grantor was being referenced.

adam-vessey:
Ping, @Islandora/7-x-1-x-committers.

DiegoPino:
@adam-vessey @bryjbrown in the meantime, what about providing the ¨corrected¨ XML form as a second option in the module? I agree with you all, especially on the need for consistency but i also don't think XML forms can-should dictate that completely, the fact that you can add-modify-your-own associations speaks about that our responsability is to provide the best experience possible, but we can't ensure consistency by just freezing a form in type of expecting everyone will use that form.

xml/thesis_form_ISLANDORA1350.xml
```diff
@@ -701,7 +662,7 @@
               <access>TRUE</access>
               <collapsed>FALSE</collapsed>
               <collapsible>FALSE</collapsible>
-              <description>The year this thesis was issued.</description>
+              <description>The year this book was issued.</description>
```
DiegoPino, line 665: why revert this?

adam-vessey: I hit the "revert" button on the PR... since this change was in the PR, it got included.

adam-vessey:
@DiegoPino: It's not really the form dictating it, so much as the expectation that the field is at that location by other modules in Scholar. As for including the updated form in the module... for the exact reason that you can add/modify/associate/etc your own forms, I'm not sure it makes sense to have a form in the module that allows for out-of-the-box inconsistency. Could instead have the form attached to the Jira ticket instead, as something of a hotfix for those who wish to start using the new form to install/apply, if they so desire, with a disclaimer highlighting that some things might be inconsistent while it is in use. Or, in the module but _without_ an association, so it is not available for use in ingest by default--it would have to be associated manually; however, this may just make things more difficult to manage in the future, when this hypothetical updated copy is no longer necessary and removed (leaving an association to a non-existent form in the DB?).

xml/thesis_form_ISLANDORA1350.xml
DiegoPino: ok, though we could have kept this tiny improvement. next time!

DiegoPino:
@bryjbrown are you ok with this one? feel free to push the buttons if so

bryjbrown (approved):

PR #285, opened 2018-03-23
Adds check to OpenURL PID in admin form

bryjbrown:
**JIRA Ticket**: https://jira.duraspace.org/browse/ISLANDORA-1923

Adds a check to validate the user-supplied email address used as the OpenURL PID.

# What's new?
Theres a new function in islandora_doi/includes/utilities.inc, `islandora_doi_openurl_pid_valid()`, that takes a supplied email and uses it to perform a CrossRef request on an example DOI (fake DOI thats valid according to CrossRef used for testing) and judges whether its valid based on the content-type of CrossRef's response. If the email address is registered with CrossRef, they get an XML response. If the email is not registered with CrossRef, they get an HTML error message. 

The admin form feeds the OpenURL PID supplied by the user to this function during a validation step; if `islandora_doi_openurl_pid_valid()` determines that the email address is invalid, users get an error message and the invalid email is not saved.

# How should this be tested?
On a fresh VM, go to the admin menu for Scholar and go to the "DOI Configuration" tab. Try to enter a random email address other than the default that you know isn't registered with CrossRef. It shouldn't let you do it and will give you an error message asking you if the provided email has been registered. If you register an email with CrossRef (or you can just use "bjbrown@fsu.edu" if you don't want to go through the hassle) and provide it, it should allow you to save that valid email address.

# Interested parties
@Islandora/7-x-1-x-committers


bryjbrown:
Looks like just one of the Travis builds failed due to a network error (couldn't reach packagist.org). 

Is there any way to force Travis to run again on the same commit?

DiegoPino:
@bryjbrown restarted the failed test (first time I do this while on a plane.. i think...).

modules/doi/includes/admin.form.inc
```diff
@@ -30,7 +30,48 @@ function islandora_doi_admin_form(array $form, array &$form_state) {
...
       '#default_value' => variable_get('islandora_doi_openurl_pid', 'user@example.com'),
     ),
+    'save_configuration' => array(
+      '#type' => 'submit',
+      '#name' => 'saveConfiguration',
+      '#default_value' => t('Save configuration'),
+    ),
   );
 
-  return system_settings_form($form);
```
jordandukart: Any particular reason you moved away from using `system_settings_form`?

modules/doi/includes/admin.form.inc
```diff
@@ -30,7 +30,48 @@ function islandora_doi_admin_form(array $form, array &$form_state) {
       '#description' => t('An identifier to call yourself, for the OpenURL endpoint. To use this service you first need to register for an account here: http://www.crossref.org/requestaccount/'),
       '#default_value' => variable_get('islandora_doi_openurl_pid', 'user@example.com'),
     ),
+    'save_configuration' => array(
+      '#type' => 'submit',
+      '#name' => 'saveConfiguration',
```
jordandukart: Shouldn't need to explicitly set a `#name` for the submit.  Also, `#default_value` is not a valid property on a submit element: https://api.drupal.org/api/drupal/developer!topics!forms_api_reference.html/7.x#submit. May be rendered moot by my comment below.

jordandukart (changes requested):

adam-vessey:
Just a heads up: This will be somewhat conflict prone with https://github.com/Islandora/islandora_scholar/pull/280.

d-r-p:
Should this be wanted, I will be happy to assist in streamlining this PR with [#280](https://github.com/Islandora/islandora_scholar/pull/280) as soon as the requested changes have been implemented.

DiegoPino:
@bryjbrown @d-r-p there are still a few concerns from @jordandukart here to address. Let's start with that and then we can move on syncing with #280 . Thanks folks

modules/doi/includes/admin.form.inc
bryjbrown: My understanding was that having `system_settings_form($form)` at the end of the form automagically saves all the entered data to the relevant variables. I wanted to implement this check as part of a `form_validate`, so I removed that function and saved all the variables individually in a `form_submit` step.

I'm basically cargo culting from https://github.com/Islandora/islandora_usage_stats/blob/7.x/includes/admin.form.inc

modules/doi/includes/admin.form.inc
bryjbrown: Once again, just copying from https://github.com/Islandora/islandora_usage_stats/blob/7.x/includes/admin.form.inc. If thats wrong, then we should change it there, too.

bryjbrown:
@DiegoPino @adam-vessey @d-r-p I replied to the code review comments, I can clean them up if needed. After https://github.com/Islandora/islandora_scholar/pull/280 is merged I can resolve any conflicts locally and push a final version.

modules/doi/includes/admin.form.inc
jordandukart: Form validation still gets triggered with `system_settings_form`.

modules/doi/includes/admin.form.inc
jordandukart: Should get changed in usage stats as well then.

modules/doi/includes/admin.form.inc
bryjbrown: Are you recommending that I add `system_settings_form` back in and take the whole `form_submit` function out?

modules/doi/includes/admin.form.inc
bryjbrown: So I just need to remove `#name` and `#default_value` then?

jordandukart: `#default_value` becomes `#value`, `#name` can be removed.

modules/doi/includes/admin.form.inc
jordandukart: Yep the submit handler can be removed and the form validation will still trigger.

d-r-p: Just a thought: would it not be more logical to use `#element_validate` on `$form['islandora_doi_openurl_pid']`, rather than create a validator for the whole form? After all, it is only this element that is checked...

whikloj: @bryjbrown Here is an example from the core Islandora config page.
https://github.com/Islandora/islandora/blob/7.x/includes/admin.form.inc#L181-L187

jordandukart: Very true @d-r-p.

d-r-p: On the other hand, in view of #280, the part that checks against the default email address should only trigger in non-legacy mode, so a form validator might be the right thing, after all...

bryjbrown: We *could* do it through an `#element_validate`, but IMHO that would make it more complex and harder to follow than just having the validation logic right below in a `form_validate` function which is the pattern most Islandora modules already follow.

I'm not opposed to this though, just giving my opinion.

DonRichards:
@bryjbrown After the last commit it's still saying Conflicting files modules/doi/includes/utilities.inc It looks fine to me but I just wanted to point it out just incase.

bryjbrown:
@DonRichards Its because of https://github.com/Islandora/islandora_scholar/pull/280 getting merged, it introduces some differences that weren't there before. We knew this would be an issue though, I'll resolve those conflicts today.

bryjbrown:
So @d-r-p let me make sure I understand whats going on now that we're living in a post-https://github.com/Islandora/islandora_scholar/pull/280 world: 

If someone has "Use legacy retrieval mode" checked, it retrieves DOIs using the CrossRef "DOI OpenURL" and the "OpenURL PID" email address.

If they DON'T have "Use legacy retrieval mode" checked, then the "DOI OpenURL" and "OpenURL PID" don't matter at all.

Just making sure, because I'm *thinking* that I don't have to update my `islandora_doi_openurl_pid_valid()` function then since it only applies to the "legacy mode".

d-r-p:
@bryjbrown Your statements about the modes are correct, so, yes, `islandora_doi_openurl_pid_valid()` does not need to be modified for non-legacy mode.

However, the default email address (`user@example.com`) is now stored in a constant (`ISLANDORA_DOI_OPENURL_PID_DEFAULT`). Similarly for the OpenURL url. See https://github.com/Islandora/islandora_scholar/blob/7.x/modules/doi/islandora_doi.module#L8-L9. Therefore, `islandora_doi_openurl_pid_valid()` needs to be touched, after all.

Note also that, if I am not mistaken, `islandora_doi_admin_form_validate()` will trigger an error when submitting the admin form with "Use legacy retrieval mode" unchecked and an invalid (i.e. default or un-registered) email address. This should not occur, as the value of "OpenURL PID" should be irrelevant in that case - just as you pointed out.

Another thing I am wondering about: since you only check the header of the response, would it not be more appropriate to make a HEAD request rather than a full GET request (see [`drupal_http_request()`](https://api.drupal.org/api/drupal/includes%21common.inc/function/drupal_http_request/7.x) and [RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html))? Of course, one would have to test if CrossRef conforms to the specs...

bryjbrown:
@d-r-p I'm trying to test this, and I've wrapped the form validate body in a check to see if legacy mode is true, but I'm running into a problem: if you have a non-default email address set to be the OpenURL PID, legacy mode can't actually be activated. You can save the form and it will tell you that your settings have been saved, but the check in the legacy mode box will disappear. This happens on the main 7.x branch as well, so I'm thinking this is a vestige from #280. 

I think that the `#default_value` for `islandora_doi_use_legacy_mode` is getting overwritten somehow, and it likely has to do with whatever is going on in https://github.com/Islandora/islandora_scholar/blob/7.x/modules/doi/includes/utilities.inc#L16. This is a long line of code with hard to read logic, and I can't untangle what its actually doing here. My best guess is that its returning the value of `islandora_doi_use_legacy_mode` with a default value set to true or false depending on some sort of check to see if the OpenURL PID is set to the default, but that doesn't really make much sense to me and I don't understand the logical checks that are happening.

Personally, I think we should just manually set the default value to be either true or false. I don't have a strong opinion on which one.

bryjbrown:
Apologies if the above comment seems harsh or grumpy, I'm having a rough morning. :grimacing:

d-r-p:
@bryjbrown Sorry! My mistake! Please give me few hours to address this. Should I make a PR against your branch once I solved this?

adam-vessey:
Ah, yeah, missed that during review... The strict comparison against `TRUE` on the [end of it](https://github.com/Islandora/islandora_scholar/pull/280/files#diff-52bb712f031b7c5151fc17f7508f7213R16) would be throwing it off.

Could just cast the return to a boolean, instead of a comparison?

d-r-p:
Correct @adam-vessey! The strict comparison should be against the integer `1`, thereby returning a boolean.

adam-vessey:
@d-r-p: Just bear in mind: Without changing how the default resolves, a strict comparison against `1` could fail, since the present comparison with the value of the PID _would_ be a boolean `TRUE`... Meaning the strict comparison with `TRUE` would pass, while a strict comparison with `1` would lead to unexpected results.

bryjbrown:
@d-r-p No worries, you can make a PR against my branch, or you could make a PR against 7.x that we could quickly merge and I could fold into my PR. Or if you want to go fast as possible, just post the updated code here in the comments and I can copy paste. Your call.

bryjbrown:
@d-r-p Whatever you do though, I would highly recommend breaking this across a few lines to make the logic of what is happening clearer, and possibly add some comments. Anything to make it absolutely clear what is happening and why.

d-r-p:
@adam-vessey You are absolutely right! I was a bit too fast here... 
@bryjbrown I will do as requested. Just give me few hours...

bryjbrown:
Huge thanks to @d-r-p for dealing with this complicated hiccup, the new `islandora_doi_use_legacy_mode()` function now not only works better, but is also far easier to read and understand thanks to breaking it up over multiple lines and lots of comments. There are also several other changes that were introduced as a PR to my PR branch (see them all here: https://github.com/bryjbrown/islandora_scholar/pull/2).

I have tested the new code and it seems to be working fine and only fires conditionally when legacy mode is active (or attempting to activate). 

I think all thats left is the code review. I incorporated the feedback @jordandukart suggested (getting rid of the form submit step and replacing it with a call to `system_settings_form()`) so we should be good on that front. I'd like to get this merged in before the code freeze tomorrow evening.

modules/doi/includes/utilities.inc
```diff
@@ -88,7 +98,38 @@ function islandora_doi_get_doi_name_from_url($doi_url) {
 }
 
 /**
- * Perform a request to a metadata provider for the given ID.
```
d-r-p, line 91: @bryjbrown [This deleted comment line](https://github.com/Islandora/islandora_scholar/pull/285/files#diff-52bb712f031b7c5151fc17f7508f7213L91) should actually replace [this newly inserted comment line](https://github.com/Islandora/islandora_scholar/pull/285/files#diff-52bb712f031b7c5151fc17f7508f7213R132) (now, CrossRef is potentially only one of many metadata providers). It seems that the update got reverted during merging of Islandora#280.
(Recap: `s/CrossRef/a metadata provider/` on [L132](https://github.com/bryjbrown/islandora_scholar/blob/cef53684998ed85ffdf02378236cefda942dc8fb/modules/doi/includes/utilities.inc#L132) of `modules/doi/includes/utilities.inc`)

bryjbrown: :+1:

DiegoPino:
@bryjbrown are @jordandukart´s concerns are addressed? Let me know. Let´s try to get this into the release!

bryjbrown:
@DiegoPino Yep, I'd say so. His two issues with the code were:
- Removal of `system_settings_form()` (which I added back)
- Wrong keys on the form submit button

I took out the submit button and added `system_settings_form()` back in, so there should be no issues now.

DiegoPino:
@bryjbrown since some folks are way today and we just moved code freeze to next week, let's wait until Monday and @jordandukart is back. I will test in the meantime too.

d-r-p:
@DiegoPino From what I understood, I would confirm that @jordandukart's issues are now addressed. Perhaps @jordandukart or @adam-vessey would also like to have a quick look at the additional changes stemming from bryjbrown#2 (in particular the "[Small improvements](https://github.com/Islandora/islandora_scholar/pull/285/commits/9fc95e848efe43accdbfb73e755bd7b2f980f9f1)" commit)? Thank you!

DiegoPino:
Cool @d-r-p . @adam-vessey or @jordandukart any chance you could give this a look to see if your concerns are addressed? We still have 48 hours left for code freeze and seems like this one is semi-required to get #280 work properly (and if not, well its a good one anyway)? Thanks folks!

modules/doi/islandora_doi.install
```diff
@@ -0,0 +1,41 @@
...
+ */
+function islandora_doi_update_7100() {
+  $variables = array(
+    'islandora_doi_openurl' => array(
+      'old_default' => 'http://www.crossref.org/openurl',
+      'new_default' => ISLANDORA_DOI_OPENURL_DEFAULT,
+    ),
+    'islandora_doi_openurl_pid' => array(
+      'old_default' => 'user@example.org',
+      'new_default' => ISLANDORA_DOI_OPENURL_PID_DEFAULT,
```
adam-vessey: Very minor, and can probably be ignored, but: Update hooks are a special thing... normally, it best to avoid using references to data outside of them, and have all values explicitly contained in the update function; for example, if we want to change the defaults again in the future, we would presumably roll another `hook_update_N()` to similarly update like we do here; however if an installation was to update across both hooks, the state when entering the second update hook may be unexpected... it would already contain the new values. This is much [more of an issue with database/schema management things](https://www.drupal.org/node/150220), but seems similar enough that I thought I should at least mention it.

adam-vessey (commented):
Typification bit looks fine, update code could be a little safer, but is probably fine.

modules/doi/islandora_doi.install
DiegoPino: @adam-vessey so you mean, don't use the constant here, just put the new default manually, right?

adam-vessey: Yeah, both here and [above](https://github.com/Islandora/islandora_scholar/pull/285/files#diff-918824dec4ab3033cac4e6b3ac88b1daR28), to have them explicitly as strings instead of using the constants.

d-r-p: Good point @adam-vessey! Thank you! @bryjbrown would you like to change this? The part with the default PID can then be taken out as it is superfluous anyway (I wrote it only for symmetry).

One question, though: if we were to put a second update function in here with the now current values as literals and the new values as constants (like in the first function), a user who already executed the first one would execute the second one just fine, right? But if they did not, the first execution would update the variables directly to the current constants (which is the concern here, correct?). But the second update hook would then not do anything, no? The retrieved value would be the `'new_default'`, and the `foreach` clause would not do any work. True, this is wasted computation time, but the result is the one expected. Or am I missing something here?

adam-vessey: No, you've got it @d-r-p... this is why I said it's very minor and can probably be ignored. The only way it would be an issue if some _other_ logic which depended on the existing value was being performed which is... entirely hypothetical. I can't really think of a case for it being necessary for this in particular, but this is not to say that none exists.

DiegoPino: @adam-vessey @bryjbrown @d-r-p i think i read somewhere (probably the hook_update_N page) that its good practice to remove old hook_update_N implementations from time to time (like keep the latest 2 old ones), which means, i guess, that we can not depend too much on them being an incremental fix, but be as standalone as possible. Right?

adam-vessey: @DiegoPino: Not... quite. From [`hook_update_N()`](https://api.drupal.org/api/drupal/modules!system!system.api.php/function/hook_update_N/7.x):

> A good rule of thumb is to remove updates older than two major releases of Drupal.

They're appear to be talking about major versions of _Drupal_, so if we had modules with update hooks from Drupal 4 era, we would/should have removed them for Drupal 7... or such is my understanding of this.

DiegoPino: hu! @adam-vessey++ true. Weird... can't think of anything 4 or 5 or 6 era that still works on 7! maybe they were talking about non-semantic versioning, where a major release could be any other subversion. (like 7.4 to 7.5?). Who knows..!

DiegoPino:
@adam-vessey @jordandukart @bryjbrown @bondjimbond let me know what the state of this one is? still on time for code freeze if you approve/address changes today. Thanks a lot, i know you are all bussy, sorry for being "pushy", i mean well!

bryjbrown:
@DiegoPino It looks like the only thing left to address is @adam-vessey's suggestion of swapping out constants with literals, right? I can try to take care of that by the end of the day. The only other issue is @jordandukart's review, which should either be resolved or dismissed.

modules/doi/islandora_doi.install
bryjbrown: @DiegoPino @adam-vessey Okay, so I was out yesterday and I'm just now getting to this and reading through. I'm not totally sure I understand whats going on here, if we change the "new_default" constant to its string literal value would it just be the exact same thing as "old_default"? For example, instead of 
```
  $variables = array(
    'islandora_doi_openurl' => array(
      'old_default' => 'http://www.crossref.org/openurl',
      'new_default' => ISLANDORA_DOI_OPENURL_DEFAULT,
    ),
    'islandora_doi_openurl_pid' => array(
      'old_default' => 'user@example.org',
      'new_default' => ISLANDORA_DOI_OPENURL_PID_DEFAULT,
    ),
  );
```
we would instead have
```
  $variables = array(
    'islandora_doi_openurl' => array(
      'old_default' => 'http://www.crossref.org/openurl',
      'new_default' => 'http://www.crossref.org/openurl',
    ),
    'islandora_doi_openurl_pid' => array(
      'old_default' => 'user@example.com',
      'new_default' => 'user@example.com',
    ),
  );
```
I'm thinking that @d-r-p's original intention was to future-proof the module from changes to the defaults by having them get overridden by whatever the current value of the constant is. If we throw out the constants and just use their current string literal values, it defeats the purpose and simply overrides their value with the exact same value. 

Another option would be to just shelve this whole update function until the defaults actually change. Right now I'm not seeing the need.

DiegoPino:
@bryjbrown sounds good. Let me know if you need a review on that and i will ping @jordandukart (ping) and aks him about that review. We can always bend the 24-hour rule a bit tomorrow if this takes more time and we have a few 👍 to support this since the previous code was already reviewed. But let's deal with that when the time comes (or runs out). -Thanks

modules/doi/islandora_doi.install
DiegoPino: @bryjbrown so... old and new are the same? and why do we need the update hook then??? ❓ 😬

adam-vessey: @bryjbrown If the values are not changing, what's the point of this update hook at all, independent of whether the values are explicitly inline or referenced via a constant?... That said, it _does_ appear that the [`islandora_doi_openurl` value is being changed](https://github.com/Islandora/islandora_scholar/pull/285/files#diff-1c1e5251eca683f1ba22e7afb4f2a63cR8)... the TLD mixup which was fixed in  https://github.com/Islandora/islandora_scholar/pull/285/commits/fb4b7d6e7304cca1d0edf5be72d4e737085dcc35 was slightly confusing the issue...

d-r-p: Thank you, @bryjbrown, for spotting the wrong tld!

Regarding your question: as far as I understood, the update hook implementation should read (although it could be simplified):
```php
/**
 * Update variable defaults.
 */
function islandora_doi_update_7100() {
  $variables = array(
    'islandora_doi_openurl' => array(
      'old_default' => 'http://www.crossref.org/openurl',
      'new_default' => 'https://www.crossref.org/openurl',
    ),
  );
  foreach ($variables as $variable => $info) {
    $value = variable_get($variable, $info['new_default']);
    if ($value == $info['old_default']) {
      variable_set($variable, $info['new_default']);
    }
  }
}
```
instead of:
```php
/**
 * Update variable defaults.
 */
function islandora_doi_update_7100() {
  $variables = array(
    'islandora_doi_openurl' => array(
      'old_default' => 'http://www.crossref.org/openurl',
      'new_default' => ISLANDORA_DOI_OPENURL_DEFAULT,
    ),
    'islandora_doi_openurl_pid' => array(
      'old_default' => 'user@example.com',
      'new_default' => ISLANDORA_DOI_OPENURL_PID_DEFAULT,
    ),
  );
  foreach ($variables as $variable => $info) {
    $value = variable_get($variable, $info['new_default']);
    if ($value == $info['old_default']) {
      variable_set($variable, $info['new_default']);
    }
  }
}
```
The purpose of this update is to use https instead of http when querying CrossRef. As I said earlier, I had included the OpenURL PID update only for symmetry.

Apologies for the tld mix-up from which the confusion seems to stem.

bryjbrown: @adam-vessey Yes, I think it was just a typo.

bryjbrown: @DiegoPino @adam-vessey So... :+1: to getting rid of this update hook?

@d-r-p Please chime in if we are misunderstanding the purpose of this update hook.

bryjbrown: @d-r-p Okay, didn't catch the s/http/https part on here, that is a valid update. I'll patch in your first example in an incoming commit. :+1:

d-r-p: ~@bryjbrown no, we need the update hook, or users who set their OpenURL PID will continue to use http for querying CrossRef. I think what I wrote above should do it...~ (scratched, for this was a reply to a now deleted comment)

bryjbrown: @d-r-p Check https://github.com/Islandora/islandora_scholar/pull/285/commits/c9b85c976aa553de6078631f2e7456b5f8398776 to verify that this is what it should look like.

d-r-p: Looks good to me! Thank you @bryjbrown!

@adam-vessey do you agree, as well?

adam-vessey: Yeah, looks good here.

bryjbrown: Awesome! I think we've finally driven a stake through this one's heart, @DiegoPino :champagne:

DiegoPino:
@jordandukart ping pong ping. Could you check on your requested changes? @bryjbrown addressed all of them and we need you to approve (if you feel all is ok) to merge. Thanks a lot!

jordandukart:
Will look at this now.

jordandukart (approved):

DiegoPino:
@jordandukart thanks man! @bryjbrown @d-r-p thanks so much for all coding and addressing concerns and @adam-vessey for the good reviews!

PR #297, opened 2018-07-05
Citeproc V2 fix

jonathangreen:
## What does this Pull Request do?

During code review of https://github.com/Islandora/islandora_scholar/pull/291 a function as removed. One reference to this function remains in the module, and it prevents the admin form from loading, if the variable isn't set.

It should be set since it gets set in hook_init and hook_install, but it is possible to load the form with the variable not set and get an error.

## How should this be tested?

* Make sure variable isn't set, load form, see error. 
* Apply patch, load form, no error.

## Interested parties
@DiegoPino

### DIFF
--- a/modules/citeproc/includes/admin.form.inc
+++ b/modules/citeproc/includes/admin.form.inc
@@ -23,7 +23,7 @@ function citeproc_admin_form() {
   $form['citeproc_backend'] = array(
     '#type' => 'radios',
     '#title' => t('What version of citeproc library should be used?'),
-    '#default_value' => variable_get('citeproc_backend', citeproc_default_backend()),
+    '#default_value' => variable_get('citeproc_backend'),
     '#options' => $options,
     '#description' => t('V2 of citeproc-php is recommended when your PHP version supports it.'),
   );
--- a/modules/citeproc/includes/admin.form.inc
+++ b/modules/citeproc/includes/admin.form.inc
@@ -23,7 +23,7 @@ function citeproc_admin_form() {
   $form['citeproc_backend'] = array(
     '#type' => 'radios',
     '#title' => t('What version of citeproc library should be used?'),
-    '#default_value' => variable_get('citeproc_backend'),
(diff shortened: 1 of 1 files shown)


DiegoPino:
@jonathangreen cool. If in hook_init and hook_install it should be set as soon as the module is enabled, but you could eventually remove the variable manually via drush and end in this situation? Is that the way to trigger the error? Or are there other less hack-ish ones?

jonathangreen:
I unset the variable with drush while I was testing, and thats what triggered the error for me.

DiegoPino:
@jonathangreen just to be totally fool proof on this, should that variable_get still keep one of the two options (without the function, just the value?) as a default?

jonathangreen:
Just doing slightly more testing, having the function there might error out the admin form even if the variable is set, this patch is still the fix though.

jonathangreen:
It can, I'm not sure which to chose as the default though. 

This is the logic we use in hook_install to determine which to use: 
https://github.com/Islandora/islandora_scholar/blob/7.x/modules/citeproc/citeproc.install#L30-L34

DiegoPino:
I do like that
```PHP
$default = citeproc_v2_supported() ? CITEPROC_BACKEND_V2 : CITEPROC_BACKEND_V1;
```
logic

jonathangreen:
Sounds good. I updated the form to use that logic to pick a default.

DiegoPino:
@jonathangreen once travis passes i will just do a quick test and proceed to "official" merge. In case of this type of bugs, that slipped under the door and that required a quick fix, we can get maybe a few additional +1 to merge before 24 hours (exceptional official), or we can wait 24 hours if that is Ok with you. I'm more for an as soon as possible scenario here.

jonathangreen:
Either way is fine with me @DiegoPino, I don't mind waiting the official 24 hours.

DiegoPino (approved):

DiegoPino:
@DonRichards  did you test? Thanks

DonRichards:
@DiegoPino Oh sorry, I should have noted that. Yes, I tested but fear I may not have fully checked everything (only because I'm not sure everything Citeproc might touch).

jonathangreen:
Thanks @DonRichards!

PR #295, opened 2018-06-26
Islandora 2251

DonRichards:
**JIRA Ticket**: https://jira.duraspace.org/browse/ISLANDORA-2251

https://docs.google.com/spreadsheets/d/1PtK16Fo2q4jTMgz17DYmV1nauvoGDEi6f2fhl9K1ZIA

# What does this Pull Request do?

Makes the XPaths editable through an admin GUI

ISLANDORA SCHOLAR XPATH CONFIGURATION

# What's new?

Islandora Scholar XPath Configuration Section located on the /admin/islandora/solution_pack_config/scholar page. It allows you to modify the previously hard coded XPaths and reset them at any time. 

# How should this be tested?
* Start islandora_vagrant and ingest an item (prior to switching to this PR)
  * Inspect it and test that the search results look like expects
  * from the terminal
```terminal
      $ vagrant ssh
      $ cd /var/ww/drupal
      $ drush vget islandora_scholar_default_search_xpath
      No matching variable found.
```
* Switch to this PR
  * Go to /admin/islandora/solution_pack_config/scholar page and click 'Save Configuration'
  * from the terminal
```terminal
      $ vagrant ssh
      $ cd /var/ww/drupal
      $ drush vget islandora_scholar_default_search_xpath
...
```
  * This time you should see several results
  * Ingest an item
  * Try modifying an XPath (easiest is to switch title with something else)
  * Check that the change shows up on a new ingest and an existing one. 
  * Also try adding new values to the citation_form.xml or thesis_form.xml and see if pointing to the new values works. 

# Additional Notes:

* This change requires documentation to be updated.
* This should not impact existing installations.
* This does **NOT** modify the submission form.
* This allows for the 52 XPaths and 3 subpaths to be set
* This exclude subpaths in a few spots where they occurred.
    * Check google doc for excluded some paths

# Interested parties
Tag (@ mention) interested parties or, if unsure, @Islandora/7-x-1-x-committers

PR #265, opened 2017-04-26
Patch 1

krebeck:
This pull request adds functionality improved on and referenced in this discussion:

https://groups.google.com/forum/#!topic/islandora/EwWeXNJ-dlk

This code adds configuration for a simple catalog search text to appear beneath the Google Scholar search text.  It is configurable in much the same way and also requires changes to utilities.inc in this directory.

It provides a way to query a discovery layer without needing an outside block or additional modules.

# How should this be tested?

This can be tested by adding the required configurations from a known good discovery layer and then testing from a citation object.

# Additional Notes:
This should not require any substantive change to the documentation.  It's pretty self-explanatory.  I improved my first draft so that you can also customize the search parameters, but it provides a good set of defaults for those using the popular Summon platform - that should cover ~30-50% of academic libraries.  So it may not work out of the box for everyone, but is easily configurable for everyone.

PR #299, opened 2018-07-17
Islandora 2251 Resubmitted

DonRichards:
## Resubmitted from https://github.com/Islandora/islandora_scholar/pull/295 due to an issue with the branch. 

**JIRA Ticket**: https://jira.duraspace.org/browse/ISLANDORA-2251

https://docs.google.com/spreadsheets/d/1PtK16Fo2q4jTMgz17DYmV1nauvoGDEi6f2fhl9K1ZIA

Makes the XPaths editable through an admin GUI

ISLANDORA SCHOLAR XPATH CONFIGURATION

### Enabled
Complete

# What's new?

Islandora Scholar XPath Configuration Section located on the /admin/islandora/solution_pack_config/scholar page. 

1. It allows you to modify the modules' XPaths.
2. It allows you reset the xpaths back to the original values.
3. It will show in real time preview of the values of a given MODS XPath value for a given PID. Preview returns the value in 1 of 4 states.
     a. Path cannot be blank (User provided XPath is blank and not allowed)
     b. Error (User provided XPath has a syntax error)
     c. Nothing found at the specified XPath
     d. Found value at the specified XPath

# How should this be tested?
* Start islandora_vagrant and ingest an item (prior to switching to this PR)
  * Inspect it and test that the search results look like expects
  * from the terminal
```terminal
      $ vagrant ssh
      $ cd /var/www/drupal
      $ drush vget islandora_scholar_default_search_xpath
      No matching variable found.
```
* Switch to this PR
  * Go to /admin/islandora/solution_pack_config/scholar page and click 'Save Configuration'
  * from the terminal
```terminal
      $ vagrant ssh
      $ cd /var/www/drupal
      $ drush vget islandora_scholar_default_search_xpath
...
```
  * This time you should see several results
  * Ingest an item
  * Try modifying an XPath (easiest is to switch title with something else)
  * Verify the MODS validation is working as expected(give it a valid value, an invalid value, a blank value and possible an error causing value). 
  * Check that the change shows up on a new ingest and an existing one. 
  * Also try adding new values to the citation_form.xml or thesis_form.xml and see if pointing to the new values works. 

# Additional Notes:

* This change requires documentation to be updated. Only placeholders were added. 
* This should not impact existing installations.
* This does **NOT** modify the submission form.
* This allows for the 52 XPaths and 3 subpaths to be set
* This exclude subpaths in a few spots where they occurred.
    * Check google doc for excluded some paths

Using some feedback (from committers call) I set the menu to default to collapsed to keep the screen clean.

# Interested parties
Tag (@ mention) interested parties or, if unsure, @Islandora/7-x-1-x-committers


DonRichards:
Here is what the entire menu looks like. 
download 1

bryjbrown:
@DonRichards 10 bajillion apologies for dragging my feet on reviewing this. :weary: :pray: 

So first things first: the most extreme of kudos once again for taking this on and getting it started. So far, it looks pretty good. I'm not ready to do a deep code read yet to verify that all possible xpaths have been variablized (probably not a word, but you know what I mean), but theres room for that later. Lets start with a top level view and work our way down.

I originally thought that this should be its own admin menu somewhere in the "Utility Modules" section, and I also thought we should limit the amount of XPaths that we actually make configurable since theres no sane reason someone would want to override the default xpath for MODS `<title>`... right?

I was wrong about that. I think on an earlier call you pointed out that sometimes there may be multiple titles, so you want to make sure you pick up the primary one and may want to augment the xpath to search for a title with a special attribute. Thats a great point that I hadn't thought of, and it also shows how flexibility trumps a narrow minded person's view of "common sense". I applaud this deep level of configurability.

As for the "make this a standalone admin menu in the 'Utility Modules' menu" part, that was wrong too. since these XPaths are being used across multiple submodules of Scholar, it seems correct to have them as part of the Scholar admin menu. Making them collapsible is also a nice touch to reduce the brain-melting cognitive load of administering Scholar, but I think making the whole thing its own tab, similar to how "DOI CONFIGURATION" and "PDF CONFIGURATION" already are, would be an even better level of organization. "XPATH CONFIGURATION" would fit the pattern nicely.

bryjbrown:
Also, it appears that Travis does not like your use of escaped single quotes. Should be an easy fix.

DonRichards:
OK, working on moving to its own tab now.

DonRichards:
OK, made some changes. I'm not sure how to check if everything is behaving as expected. Any thoughts?

DonRichards:
Updated the screen shots to reflect these changes

DonRichards:
@bryjbrown Looks like it's ready for some testing and a deeper dive. I've changed everything I can see needed to be changed based on your feedback. I believe the documentation can be added to at a later time by someone with more knowledge of the impact of the values but the placeholders are in place for them.

bondjimbond:
Testing this out now... getting some strange results. 

Created a simple Thesis object. 7.x gives me the following citation:

> Weigel, B. (2008). My Thesis: Subtitle. doi:3301.399/bweigel33

Checked out ISLANDORA-2251. Played around a bit with xpaths for Genre only. New result: title appears twice now...

> My Thesis: Subtitle. (2008). My Thesis: Subtitle. doi:3301.399/bweigel33

Clicked the "reset to defaults" button, confirm that the changes I made are gone. Result: still has two titles.

Check out 7.x again: back to one title.

DonRichards:
@bondjimbond Thanks for that. I'll look into it. I don't believe I changed any logic so I'm not completely sure how this behavior was introduced.

DonRichards:
@bondjimbond When you hit reset, did you also hit save afterwards? I believe I have an issue with the save button. It's re-saving the reset values. Just hitting reset alone should drop everything back to the original defaults.

DonRichards:
OK, I see what I did. I combined the 2 values (didn't realize they were different and thought it was an irregularity in the code). This was the only time I did this. I'm adding it back.

DonRichards:
@bondjimbond would mind mind checking again when you have a moment? I believe I was able to iron out all of the irregularities.

DonRichards:
Didn't fix it. Not sure why it's replicating the title & subtitle. Looking through the code now.

d-r-p:
Not sure if this helps with the issue, @DonRichards, but it seems to me that `'islandora_scholar_xpaths_title_subtitle'` (by the way, I would have called the new one `'islandora_scholar_xpaths_subtitle'` instead of `'islandora_scholar_xpaths_title_sub_title'`, but that is your call) was not the only variable that was re-used inappropriately - see, e.g., `'islandora_scholar_xpaths_local_author'` on [L519](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-dc3682c7ee8afb241d9961a4b93eb82eR519) vs. [L602](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-dc3682c7ee8afb241d9961a4b93eb82eR602).

DonRichards:
## Some unresolved conflicts I needed sorted out before I can complete. 
Which of these are the correct MODS values for the field?
Listed is the file/line # where the value is declared. 
### Abstracts
islandora_scholar/includes/coins.inc:**162**:  Abstract = **'//mods:abstract'**
islandora_scholar/modules/citeproc/includes/converter.inc:**127**:  Abstract = **'//mods:mods[1]/mods:abstract'**
islandora_scholar/modules/citeproc/citeproc.api.php:**23**: Abstract = **'//mods:mods[1]/mods:abstract'**
### Genre
islandora_scholar/modules/islandora_google_scholar/islandora_google_scholar.module:**110**: Genre = **'//mods:genre'**
islandora_scholar/includes/coins.inc:**184**: Genre = **'//mods:genre'**
islandora_scholar/modules/citeproc/includes/converter.inc:**178**:  Genre = **'//mods:mods[1]/mods:genre'**
### Issue
islandora_scholar/includes/coins.inc:**132**: Issue = **'//mods:part/mods:detail[@type="issue"]/mods:number'**
islandora_scholar/modules/citeproc/includes/converter.inc:**138**: Issue = **'//mods:mods[1]/mods:part/mods:detail[@type="issue"]/mods:number'**
islandora_scholar/modules/islandora_google_scholar/islandora_google_scholar.module:**148**: Issue = **'//mods:part/mods:detail[@type="issue"]/mods:number'**
### Volume
islandora_scholar/includes/coins.inc:**127**:  volume = **'//mods:part/mods:detail[@type="volume"]/mods:number'**
islandora_scholar/modules/islandora_google_scholar/islandora_google_scholar.module:**130**: volume = **'//mods:part/mods:detail[@type="volume"]/mods:number'**
islandora_scholar/modules/citeproc/includes/converter.inc:**137**: volume = **'//mods:mods[1]/mods:part/mods:detail[@type="volume"]/mods:number'**

DonRichards:
@d-r-p Thanks. That was helpful. Assuming it passes Travis I believe this is good to test now. @bondjimbond @bryjbrown

bryjbrown:
@DonRichards Regarding your question about which XPath default values to use, you can see that all the XPaths are the same, they just sometimes have a preceeding `//mods:mods[1]/` in front of them.

In my humble opinion these are the better XPaths to choose simply because they are more specific, they are finding the element in question that is a direct child of the root `mods:mods` element. The other XPaths that don't start with `//mods:mods[1]/` are technically looking for ANY instance of those elements, which could possibly match subelements for a related item, for instance. This has a much higher chance of happening for fields like `mods:title`, and even though these fields are not likely to get reused in a way that isn't describing the primary record, it would still be better from a consistency standpoint to have every default prefaced with `//mods:mods[1]/`. 

I've ranted about this before in https://github.com/Islandora/islandora_scholar/pull/265#issuecomment-363545275 so this isn't my first time making this case.

@Islandora/7-x-1-x-committers with XPath experience feel free to chime in with opinions about the relative merits of using either.

DiegoPino:
+1 to @bryjbrown comment
Since `//`means relative path, xpaths starting with that will be resolved independently of a previous parent element. In that context, i do feel using `mods:mods[1]` at the beginning is important to make sure you are targeting the right mods record in a [modsCollection](https://www.loc.gov/standards/mods/v3/mods-3-0-outline.html#collection_root) as root element situation, where many `mods:mods` records can exist (or at least one!), but also because it forces you to be explicit about which and where of the many possible named equally elements you are targeting. `<title>` is one popular tag that needs disambiguation but basically any other that can be contained by multiple different ones.

Question: what happens here if a user provided XPATH is malformed? Will it just fail silently or will it explode with some php warning? How can people test this? Could we provide a PID so settings can be tested against it? Too much to ask?

DonRichards:
@bryjbrown Thanks for that. I'll plan on making those changes but I'll keep an eye out for feedback (incase of an issue with it). 
@DiegoPino I'm not sure how complex this PR will get if I add a validation process on top. I agree it probably could be done and would be useful.

bryjbrown:
@DonRichards @DiegoPino It looks like you could use `DOMXPATH::query($xpath)` to quickly test if an XPath is valid or not (http://us2.php.net/manual/en/domxpath.query.php). Given an arbitrary XPath, it should return `FALSE` if invalid. You could set up a form_validate step for the admin form with all the XPaths and loop through each one checking for validity. 

Credit to https://stackoverflow.com/questions/25919087/validate-xpath-syntax-correctness-in-php for this solution.

DonRichards:
Putting a note here from Committers' call; Diego suggested having a place to input a PID to show what the values would look like in real time.

DonRichards:
@bryjbrown do you have an example of the code your were suggesting at the interest group call? DOMXPATH code?

DonRichards:
@bryjbrown The more I think about it, the more I like @DiegoPino 's idea about a sample PID as a way to validate instead. My thoughts is a simple Input fields for a PID and a it calls the thumbnail next to the input to show it's a valid PID. I think the code might be just as complicated. Thoughts?

DonRichards:
@DiegoPino I'm having some issues with testing that. Any ideas how to load an image (ajax) into the form without reloading the page from a input for the PID? Big question, sorry.

DiegoPino:
@DonRichards sorry busy day, but can get you a hand tomorrow if needed. Curious: why an image upload? You need a form that takes a PID, on Ajax submit checks if that PID can be loaded as an Islandora object and it's MODS datastream can be accessed, if so, that MODS is read into a DOm document and the xpaths are applied one after the other keeping an array of success and failures that can be displayed once finished. Or have you something fancier in mind?

DonRichards:
This is really how I was thinking of your suggestion for the initial input and render the thumbnail to confirm it loaded. This isn't a working example.

DiegoPino:
Fancy! Well in that case it is not that complicated. Just take in account that all the logic always happens in the form builder function itself, so loading the PID, loading the TN etc are all actions that happen based on conditionals (like "did someone submit a pid ? == form_state['values']['daPID']"), and the ajax part is really just a hint for drupal so it knows on ajax event and after that happened, what piece to return to replace via ajax an existing form chunck..


So basically:

Inside  function islandora_scholar_admin_form()  you add a check the submitted PID and if so, you load the object, theme an img with the thumbnail there and you load the whole object´s MOD. That will run even if ajax is not enabled. Then you add ajax capabilities to that ajax field and blur or change? which will call an ajax callback that will return the piece of the form that needs to be updated, in this case your whole form really.

Does this make sense? i can share code or make a pull against yours tomorrow

DonRichards:
```php
  $form['islandora_scholar_xpaths_demo'] = array(
    '#type' => 'textfield',
    '#title' => t('Demo Object'),
    '#description' => t('PID to check these values against.'),
    '#states' => array(
      'visible' => array(
        ':input[name="islandora_scholar_xpaths_enabled"]' => array('checked' => TRUE),
      ),
    ),
    '#default_value' => '',
    '#autocomplete_path' => 'islandora_compound_object/autocomplete/child',
    '#ajax' => array(
      'page callback' => 'islandora_scholar_xpaths_ajax_get_ajax',
      'page arguments' => array(3),
      'type' => MENU_CALLBACK,
      'access arguments' => array('access content'),
      'delivery callback' => 'islandora_scholar_xpaths_ajax_callback',
      '#prefix' => '<div>',
      '#suffix' => '<img src="/islandora/object/ $form['islandora_scholar_xpaths_demo']/datastream/TN/view"></a></div>',
    ),
````
Something like this but not completely sure how to make it work with the input to concatenate into a string with an `</a>`

**Comment by DonRichards** [comment]
Progress! Kinda. It's automatically filling in the input with an object's PID and once selected (or deselected) it loads the thumbnail as a way to confirm the PID is valid. I was recording this from my laptop (super slow). Apologies. 
untitled

**Comment by DonRichards** [comment]
OK, it's working, ready for testing and should behave as expected. @bryjbrown @DiegoPino @bondjimbond 
Complete

**Comment by bondjimbond** [comment]
@DonRichards I should be able to find some time to test today or later this week. I'll wait till Travis is satisfied and then take a look.

**Comment by DonRichards** [comment]
Thanks @bondjimbond

**Comment by DonRichards** [comment]
I was not using the correct phpcs parameters within my IDE. I seriously need to configure my IDE to match TRAVIS/PHPcs linting. Manually ran the checks to verified everything was corrected. Sorry for all of the commits.

**Comment by bondjimbond** [comment]
@DonRichards Starting to test now. On 7.x an example citation displays normally. On this branch, there are problems with the citation displaying the title twice.

If I check the "Custom" checkbox and click Save without changing anything, I get the double title issue, which leads me to believe there must be something wrong in the defaults...?

Here's the really weird behaviour... If I click "Reset all Xpaths to defaults" and close the overlay, I get my regular citation again -- good. But if I click the reset button, which un-checks the "custom" box, AND THEN click the Save button, I get the double title problem. It looks like the Save button is saving the custom entries even though the "custom xpaths" box is not checked.

**Comment by bondjimbond** [comment]
Regarding the PID-checking function: awesome feature, but it's not working on my machine.
screen shot 2018-08-08 at 10 52 05 am

**Comment by DonRichards** [comment]
As for your 2nd comment, did you click inside the text area then tab to the next? I guess I really should make it refresh at least once. 
screen shot 2018-08-08 at 10 52 05 am

**Comment by bondjimbond** [comment]
@DonRichards Ah, yes -- if I click into one of the custom Xpath fields and click out, it refreshes. So that works, but certainly once the initial field validates to say "Successfully loaded PID!", it should also refresh the others if at all possible.

**Comment by DonRichards** [comment]
Sounds good. I'm looking into making that happen.

**Comment by bondjimbond** [comment]
With the preview working, it shows me that there's a problem with reading the xpaths. Aside from the initial title field, every subsequent field is showing either the subtitle, or nothing.

screen shot 2018-08-08 at 2 55 51 pm

This would partially explain the issue of repeating information in the citation.

File: includes/admin.form.inc [hunk 1]
```diff
@@ -351,22 +354,1176 @@ function islandora_scholar_admin_pdf_form($form, &$form_state) {
...
+    'islandora_scholar_xpaths_event_property',
+    'islandora_scholar_xpaths_family',
+    'islandora_scholar_xpaths_genre',
+    'islandora_scholar_xpaths_given',
+    'islandora_scholar_xpaths_host_title',
+    'islandora_scholar_xpaths_ISBN',
+    'islandora_scholar_xpaths_issn',
+    'islandora_scholar_xpaths_issue',
+    'islandora_scholar_xpaths_local_author',
+    'islando
```
d-r-p: Just spotted by coincidence: some HTML seems to have sneaked in here... ;)
[edit: note that I am referring to [L1361](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R1361) - GitHub attached my comment to the line above for some reason...]

d-r-p:
Could it be that in the calls to `_mods_id()` the first argument should read `$form_state['values']['islandora_scholar_xpaths_citeproc']['_variable_name_']` rather than just `$form_state['values']['_variable_name_']`?

d-r-p:
@DonRichards Just out of curiosity: any reason you switched from `t()` to `check_plain()` in the `'#description'` fields? Would that not make translation impossible? Note that, if the reason is that you wish to include contents of variables in the text, you can achieve this using placeholders and an additional array parameter - see the documentations for [`t()`](https://api.drupal.org/api/drupal/includes!bootstrap.inc/function/t/7.x) and [`format_string()`](https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/format_string/7.x).

includes/admin.form.inc
DonRichards: Thanks, I'll correct this. Wow completely missed that.

DonRichards:
@d-r-p 

- The $form_state['values']['_variable_name_'] was giving me the correct value. I hadn't tried the other. 
- The only reason I switched from t() was phpcs complained. I hadn't tried format_string. It does look like format_string might be better suited. I'll give it a go.

d-r-p:
@DonRichards My suggestion was to use `t()` - I referred you to the documentation of `format_string()` only for an explanation of the three types of placeholders.

d-r-p:
@DonRichards Regarding the other thing, could it be that @bondjimbond has `#tree` set to `TRUE` and you not? Perhaps setting `#tree` to `FALSE` explicitly solves [the issue he saw](https://github.com/Islandora/islandora_scholar/pull/299#issuecomment-411514923)?

DonRichards:
@d-r-p well t() function throws an error when TRAVIS runs because of the concatenated variables and strings. Where is the `#tree` setting?

d-r-p:
@DonRichards I mean, of course, `t()` using placeholders instead of concatenation (you can see an example [here](https://github.com/Islandora/islandora_xml_forms/blob/7.x/elements/includes/creative_commons.inc#L195))... Regarding `#tree` (see also [this](https://www.drupal.org/docs/7/api/form-api/tree-and-parents)), I _think_ it is sufficient to set the property at the root of your `$form` array in `islandora_scholar_admin_xpath_form()` (if you do not set it again more inside). However, perhaps it is better to check first, whether or not this is what is biting @bondjimbond (in my experience, messing with `#tree` can have all sorts of unexpected effects).

DonRichards:
@d-r-p @bondjimbond I believe I got this straighten out. This is ready to test.

d-r-p:
Sorry, @DonRichards, if I am insisting on this, but I feel the `'#description'` fields should really be populated using `t()` (with placeholders and additional argument to include variable contents, whenever necessary) in order to ensure translatability. Or do you (or anyone) think I am wrong about this?

includes/admin.form.inc
```diff
@@ -351,22 +354,1218 @@ function islandora_scholar_admin_pdf_form($form, &$form_state) {
...
+  ));
+  drupal_add_js("(function($) {
+    Drupal.behaviors.islandora_scholar = {
+      attach: function (context, settings) {
+      $('body').on('load', function(){
+        console.log('xxx');
+      });
+      }
+    };
+  })(jQuery);", 'inline');
```
d-r-p: Am I seeing this right that [L1314](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R1314) through [L1322](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R1322) were added for debugging purposes only?

DonRichards: Nope, that's actually for the refresh button. I didn't see a way to trigger a refresh of all of the groups without this. Using AJAX to refresh a $form['group_a'] is easy but having it also refresh $form['group_b'] didn't seem plausible.

d-r-p: I see. Thank you @DonRichards for the clarification. I was lead to the assumption above by [L1318](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R1318) (`console.log('xxx');`) - should that line be removed in order not to clutter up the user's browser-log?

DonRichards:
@d-r-p I'm not completely sure how to make a t() work with this. Is it possible to include logic within the t()? 
Looking at this code for example:
```php
...
'#description' => check_plain('Some text ' . (!empty($form_state['values']['islandora_scholar_xpaths_demo']) ? "Testing against pid: " . $form_state['values']['islandora_scholar_xpaths_demo'] : "")),
...
```
Any suggestions on how to wrap/restructure this?

DonRichards:
Sorry, I'm being ridiculous. Just tried it. Work. Uploading now.

d-r-p:
Not sure, but perhaps 
```php
...
  '#description' => t('Some text. @testing_msg', array(
    '@testing_msg' => (!empty($form_state['values']['islandora_scholar_xpaths_demo']) ? t("Testing against pid: @pid", array('@pid' => $form_state['values']['islandora_scholar_xpaths_demo'])) : ""),
  )),
...
```
<strike>or (perhaps more elegantly)
```php
...
  '#description' => (!empty($form_state['values']['islandora_scholar_xpaths_demo']) ? t("Some text. Testing against pid: @pid", array('@pid' => $form_state['values']['islandora_scholar_xpaths_demo'])) : t('Some text.')),
...
```
</strike>
could work?

[EDIT: Second option [not accepted by TravisCI](https://travis-ci.org/Islandora/islandora_scholar/jobs/416990678#L1238-L1252).]

d-r-p:
Great, @DonRichards! ~On first sight, it looks fine (although I am predicting DCS issues - sorry).~ [EDIT: see below] ~One thing, though. You are using `!placeholder` instead of `@placeholder`: did you sanitise what will be inserted using `check_plain()` or `filter_xss()` before, as suggested by the [`format_string()` doc](https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/format_string/7.x)?~ [EDIT: Ignore that, the code was there before]

[EDIT: On second sight, and seeing TravisCI's output, lines like [L390](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R390) should be modified from
```php
      '#description' => t('These XPaths are shared within most of the submodules. To see the MODS value after editing an entry, tab to the next field. ' . (!empty($form_state['values']['islandora_scholar_xpaths_demo']) ? "Testing against pid: " . $form_state['values']['islandora_scholar_xpaths_demo'] : "")),
```
to something like
```php
      '#description' => (!empty($form_state['values']['islandora_scholar_xpaths_demo']) ? t('These XPaths are shared within most of the submodules. To see the MODS value after editing an entry, tab to the next field. Testing against pid: @pid'), array( '@pid' => $form_state['values']['islandora_scholar_xpaths_demo'])) : t('These XPaths are shared within most of the submodules. To see the MODS value after editing an entry, tab to the next field.')),
```
]

bondjimbond:
Slightly better, but still getting a mess.

Here is the first section of the admin form:
screen shot 2018-08-15 at 4 05 16 pm

- Family name - No results (even though there is relevant data in the object)
- Given name - Shows the abstract instead of the given name
- Authors - Hard for me to parse this Xpath, so I can't tell if it's working as advertised or not - but it's only showing the family name and not the given.
- Created Date - Showing the abstract

d-r-p:
Sorry, @DonRichards, for being picky, but I do not think it is necessary (nor desirable, probably), to wrap HTML-only strings in `t()` ([L266](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R266), [L267](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R267), etc...)

d-r-p:
And, being overly picky, I also would like to bring to your attention the good practice suggested in https://www.drupal.org/node/322732:
> How you choose the name of placeholders is completely up to you, but it is good to keep some best practice. Try to make sure that your placeholder names are clear for translators. In the above example, using `@n` or `@usern` or `@name` or `@user` might not be as clear as `@username`. Unfortunately Drupal itself has some inconsistencies in how these placeholders are named (including the use of dashes or underscores);  you should do better.

This applies to, e.g., [L448](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R448), I would say.

d-r-p:
@bondjimbond @DonRichards Dumb question: are you both testing this PR on a vanilla (and up-to-date) vagrant box?

bondjimbond:
@d-r-p 

> Dumb question: are you both testing this PR on a vanilla (and up-to-date) vagrant box?

Yup.

DonRichards:
@d-r-p  Yes, without a doubt. I'm juggling islandora_vagrant with out in house VM. So I force myself to rebuild most days.

d-r-p:
@DonRichards @bondjimbond May I suggest, then, that you exchange MODS at this point? @DonRichards seems to have one that works, @bondjimbond one that fails.
Since it is not a configuration issue, I am quite stunned that only one of you sees a mess with the retrieved values...

d-r-p:
Sorry, @DonRichards, my bad. It seems that one cannot set `'#title'` or `'#description'` in a conditional way, even if both branches are sanitised `t()` strings... Therefore, for [L399](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R399), [L413](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R413), [L427](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R427) and [L441](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R441), I would suggest trying the first solution in [my earlier comment](https://github.com/Islandora/islandora_scholar/pull/299#issuecomment-413321749). At least, this has the benefit of not repeating identical strings.

PS: Note that there are [a few other TravisCI issues](https://travis-ci.org/Islandora/islandora_scholar/jobs/416990678#L1238-L1252), as well.

includes/admin.form.inc
```diff
@@ -351,22 +354,1231 @@ function islandora_scholar_admin_pdf_form($form, &$form_state) {
...
+  module_load_include('inc
```
d-r-p: Ideally, `$build_string` would also be populated using `t()`, as in
```php
        $build_string = t('For pid "@pid": <span style="color:blue">@mods_value_string</span>', array(
          '@pid' => $pid,
          '@mods_value_string' => $mods_value_string,
        ));
```
Similary for [L1567](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R1567), [L1570](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R1570) and [L1573](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R1573) above.

d-r-p: [EDIT: No idea what went wrong, but this comment should have been attached to [L1581](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R1581):
```php
      echo 'Message: ' . $e->getMessage();
```
]

Should we not be using
```php
      drupal_set_message(t('@message', array('@message' => $e->getMessage())), 'error');
```
or
```php
      watchdog('islandora_scholar', 'Got exception: @message', array('@message' => $e->getMessage()));
```
or something?

d-r-p: Note: ~this comment was re-issued since GitHub attached it to the wrong line...~ [EDIT: that did not work, either... For some reason, my comments all get attached to [L1333](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R1333)]

[EDIT: No idea what went wrong, but this comment should have been attached to [L1576](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R1576):
```php
        $build_string = 'For pid "' . $pid . '": <span style="color:blue">' . $mods_value_string . '</span>';
```
]

Ideally, `$build_string` would also be populated using `t()`, as in
```php
        $build_string = t('For pid "@pid": <span style="color:blue">@mods_value_string</span>', array(
          '@pid' => $pid,
          '@mods_value_string' => $mods_value_string,
        ));
```
Similary for [L1567](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R1567), [L1570](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R1570) and [L1573](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R1573) above.

bondjimbond:
> @DonRichards @bondjimbond May I suggest, then, that you exchange MODS at this point? @DonRichards seems to have one that works, @bondjimbond one that fails.
Since it is not a configuration issue, I am quite stunned that only one of you sees a mess with the retrieved values...

Good idea. Mine is attached. 
[MODS Record (3).xml.zip](https://github.com/Islandora/islandora_scholar/files/2297546/MODS.Record.3.xml.zip)

As you can see, there's nothing unusual in there, particularly nothing that should lead to a mods:originInfo/mods:dateIssued xpath pointing at mods:abstract...

DonRichards:
@d-r-p Tried I complete different approach. This was solved in a different function within the admin form already. Completely missed. So I added it and corrected the TRAVIS errors and removed a commented out line I left in mistake.

DonRichards:
@bondjimbond Does that MODS file have and show all of the field values in a clean islandora_vagrant? I wonder if there is a bug in reading the MODS files to begin with. Everything looks correct, it's just not coming up.

bondjimbond:
@DonRichards Got any advice on booting up a clean islandora vagrant without destroying my current one?

includes/admin.form.inc
```diff
@@ -360,20 +360,42 @@ function islandora_scholar_admin_pdf_form($form, &$form_state) {
  *   Drupal form definition.
  */
 function islandora_scholar_admin_xpath_form($form, &$form_state) {
+  form_load_include($form_state, 'inc', 'islandora_scholar', 'includes/admin.form');
+  if (!empty($form_state['values']['islandora_scholar_xpaths_demo'])) {
+    $islandora_scholar_xpath_pid = $form_state['values']['islandora_scholar_xpaths_demo'];
+    exec($islandora_scholar_xpath_pid, $output, $return_value);
+    $islandora_scholar_xpaths_general_description = t('These XPaths are shared within most of the submodules. To see the MODS value after editing an entry, tab to the next field. Testing against pid: !pid <img src="', array(
+        '!pid' => "$islandora_scholar_xpath_pid",
+      )) . url('misc/watchdog-ok.png') . '"/><br/><br/>';
```
d-r-p, line 368: I think it would be better to write "complete" sentences in `t()`, as in
```php
      $islandora_scholar_xpaths_general_description = t('These XPaths are shared within most of the submodules. To see the MODS value after editing an entry, tab to the next field. Testing against pid: !pid <img src="!img_url"/><br/><br/>', array(
        '!pid' => "$islandora_scholar_xpath_pid",
        '!img_url' => url('misc/watchdog-ok.png'),
      ));
```
to be easier on translators... Similarly for [L370](https://github.com/Islandora/islandora_scholar/pull/299/commits/e5216a82ff3b166d0c8c98db21689fa39c7fa525#diff-0cdd19abe14e51fe9e7f061ec72c86d3R370), [L373](https://github.com/Islandora/islandora_scholar/pull/299/commits/e5216a82ff3b166d0c8c98db21689fa39c7fa525#diff-0cdd19abe14e51fe9e7f061ec72c86d3R373) and [L376](https://github.com/Islandora/islandora_scholar/pull/299/commits/e5216a82ff3b166d0c8c98db21689fa39c7fa525#diff-0cdd19abe14e51fe9e7f061ec72c86d3R376).

Also, is it safe to use `!placeholder` (especially `!pid`, since that is unprocessed user input)?

includes/admin.form.inc
```diff
@@ -360,20 +360,42 @@ function islandora_scholar_admin_pdf_form($form, &$form_state) {
  *   Drupal form definition.
  */
 function islandora_scholar_admin_xpath_form($form, &$form_state) {
+  form_load_include($form_state, 'inc', 'islandora_scholar', 'includes/admin.form');
+  if (!empty($form_state['values']['islandora_scholar_xpaths_demo'])) {
+    $islandora_scholar_xpath_pid = $form_state['values']['islandora_scholar_xpaths_demo'];
+    exec($islandora_scholar_xpath_pid, $output, $return_value);
```
d-r-p, line 365: Forgive my ignorance, but what is achieved here?

includes/admin.form.inc
```diff
@@ -454,14 +488,13 @@ function islandora_scholar_admin_xpath_form($form, &$form_state) {
     '#type' => 'textfield',
     '#title' => t('Demo Object'),
     '#field_suffix' => (!empty($form_state['values']['islandora_scholar_xpaths_demo']) && (islandora_object_access(ISLANDORA_VIEW_OBJECTS, islandora_object_load($form_state['values']['islandora_scholar_xpaths_demo'])) == 1)) ? islandora_scholar_xpaths_demo_prefix_callback($form_state['values']['islandora_scholar_xpaths_demo']) : t("Modify to refresh value (PID for demo object is required)."),
-    '#description' => t('PID to check these values against. "@option"', array('@option' => ((!empty($form_state['values']['islandora_scholar_xpaths_demo'])) && (islandora_object_access(ISLANDORA_VIEW_OBJECTS, islandora_object_load($form_state['values']['islandora_scholar_xpaths_demo'])) == 1) ? $form_state['values']['islandora_scholar_xpaths_demo'] : ""))),
+    '#description' => t('PID to check these values against. @option', array('@option' => ((!empty($form_state['values']['islandora_scholar_xpaths_demo'])) && (islandora_object_access(ISLANDORA_VIEW_OBJECTS, islandora_object_load($form_state['values']['islandora_scholar_xpaths_demo'])) == 1) ? $form_state['values']['islandora_scholar_xpaths_demo'] : ''))),
```
d-r-p, line 482: In view of [my earlier comment](https://github.com/Islandora/islandora_scholar/pull/299#issuecomment-413333397), how about `... t('PID to check these values against: @pid.', array('@pid' => ...`?

includes/admin.form.inc
```diff
@@ -1564,21 +1597,32 @@ function _mods_id($mods_value = NULL, $pid = NULL) {
       }
       // Build the field suffix.
       if (strlen($mods_value) == 0) {
-        $build_string = 'For pid "' . $pid . '" <span style="color:red">Path cannot be blank</span>';
+        $build_string = t('For pid !pid: <span style="color:red">Path cannot be blank</span>', array(
+          '!pid' => $pid,
```
d-r-p, line 1568: Same question: is it safe to use `!placeholder`? Similarly for [L1606](https://github.com/Islandora/islandora_scholar/pull/299/commits/e5216a82ff3b166d0c8c98db21689fa39c7fa525#diff-0cdd19abe14e51fe9e7f061ec72c86d3R1606), [L1611](https://github.com/Islandora/islandora_scholar/pull/299/commits/e5216a82ff3b166d0c8c98db21689fa39c7fa525#diff-0cdd19abe14e51fe9e7f061ec72c86d3R1611), [L1616](https://github.com/Islandora/islandora_scholar/pull/299/commits/e5216a82ff3b166d0c8c98db21689fa39c7fa525#diff-0cdd19abe14e51fe9e7f061ec72c86d3R1616) and [L1617](https://github.com/Islandora/islandora_scholar/pull/299/commits/e5216a82ff3b166d0c8c98db21689fa39c7fa525#diff-0cdd19abe14e51fe9e7f061ec72c86d3R1617).

d-r-p:
@DonRichards Could you, perhaps, ingest the MODS from @bondjimbond and see how it behaves on your VM? Also, getting yours into @bondjimbond's VM would be interesting...

d-r-p (commented):
A few remarks on the latest commit only...

includes/admin.form.inc
```diff
@@ -351,22 +354,1275 @@ function islandora_scholar_admin_pdf_form($form, &$form_state) {
...
+    '#ajax' => array(
+      'callback' => 'islandora_scholar_xpaths_citeproc_callback',
+      'wrapper' => 'islandora_scholar_xpaths_citeproc_title',
+    ),
+   '#field_suffix' => (!empty($form_state['values']['islandora_scholar_xpaths_demo']) ? _mods_id($form_state['values']['islandora_scholar_xpaths_local_collection-editor'], check_plain($form_state['values']['islandora_scholar_xpaths_demo'])) : t("Modify to refresh value (PID for demo object is required).")),
+  ));
+  $form['islandora_scholar_xpaths_citeproc'] += _islandora_scholar_add_variable('islandora_scholar_xpaths_role_term', 'mods:role/mods:roleTerm', array(
+    '#type' => 'textfield',
+    '#title' => t('Role Term'),
+    '#description' => t('Default Value: mods:role/mods
```
d-r-p: (Being afraid of having my comment getting attached to the wrong line, again, let me point out that the following is relative to [L1376](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R1376).)

Maybe I am seeing this wrong, but would this line not cause _severe_ havoc if a user were to enter "rm -rf /" (or similar) as the path to `pdftotext`? I know that [this is legacy code](https://github.com/Islandora/islandora_scholar/blob/39750e5097f938bc58fe981a48ec9256f593d2ea/includes/admin.form.inc#L226-L238), but I feel we should not leave this as is...

DonRichards: OK, will do. This just came in as a complaint to the security group a few days ago so it would make sense to sanitize it.

d-r-p: @DonRichards Well, if we sanitise it here, we should probably do this in coordination with the ISRT, lest we implement a different (less powerful) solution than them that they might overlook to modify at a later stage.

DonRichards:
I believe there is an issue within how my code is referencing some relative values. There may also be an issue with scholar. This is a screenshot of a submission on a fresh vagrant up without this PR. There is missing results
Before
Just posted some updates but still working on the relative ones.

d-r-p:
@DonRichards @bondjimbond Not sure what to suggest here... Should we perhaps try to troubleshoot the original code, before we finish this PR here? It seems to me that we might have a hard time deciding if the new or the old code is at fault, otherwise...

DonRichards:
@d-r-p I believe changing the default to include //mods:mods in front of the original value. Any reason I shouldn't make this the default? You can place "//mods:mods/" prior to Family and given to make it work.

bondjimbond:
@DonRichards I think that's a good start. A few of the fields I've been having trouble with work properly when you ad //mods:mods/ to the beginning. Making that default would help.

DonRichards:
Here's a list of MODS values that aren't currently showing up. I believe it may be a lack of values  in the test MODS file and not an issue with the code. I'd prefer to know for certain before this goes in. I'm not well versed in MODS creation. Anyone have the time to modify a MODS file to include these? 

```xml
//mods:mods[1]/mods:identifier[@type="u2"]
//mods:mods[1]/mods:relatedItem[@type="host"]//mods:date
//mods:mods[1]//mods:dateOther[@type="embargo"]
//mods:mods[1]/mods:titleInfo/mods:title/../mods:nonSort
//mods:mods[1]/mods:relatedItem[@type="series"]/mods:titleInfo[not(@type)]/mods:title
//mods:mods[1]/mods:originInfo/mods:edition
//mods:mods[1][mods:genre[@authority="local"]="conferencePaper"]/mods:relatedItem/mods:titleInfo/mods:title
//mods:mods[1][mods:genre[@authority="marcgt"][text()="conference publication"]]/mods:originInfo/mods:place/mods:placeTerm
//mods:mods[1][mods:genre[@authority="local"][text()="conferencePaper"]]/mods:originInfo/mods:place/mods:placeTerm
//mods:mods[1]/mods:relatedItem[@type="series"]/mods:titleInfo/mods:partNumber
//mods:mods[1]/mods:identifier[@type="accession"]
//mods:mods[1]/mods:identifier[@type="mid"]
//mods:mods[1][mods:genre[@authority="marcgt"][text()="conference publication"]]/mods:relatedItem/mods:titleInfo/mods:title
//mods:mods[1]/mods:part/mods:extent[@unit="pages"]
//mods:mods[1]/mods:genre[@authority="endnote"]
//mods:mods[1]/mods:genre[@authority="marcgt"]
//mods:mods[1]/mods:relatedItem/mods:genre[@authority="marcgt"]
//mods:mods[1]/mods:originInfo/mods:dateOther[@type="season"]
//mods:mods[1]/mods:originInfo/mods:dateCaptured
//mods:mods[1]/mods:relatedItem[@type="series"]/mods:name
```

My example [MODS Record (2).xml.txt file](https://github.com/Islandora/islandora_scholar/files/2324940/MODS.Record.2.xml.txt)
Github blocks XML files so I renamed it to a TXT file.

bryjbrown:
@DonRichards Sorry for sleeping on this thread, Outlook has been sending all of my GitHub notifications to the junk folder. Out of sight, out of mind.

I'm all caught up now and will try to support this merge however I can. I've added most of the requested elements to the provided MODS record:
[notActuallyA.txt](https://github.com/Islandora/islandora_scholar/files/2325502/notActuallyA.txt)

However some of the XPaths you listed I just can't figure out so I'm not sure how to add an element that satisfies them, such as:
```
//mods:mods[1][mods:genre[@authority="local"]="conferencePaper"]/mods:relatedItem/mods:titleInfo/mods:title
//mods:mods[1][mods:genre[@authority="marcgt"][text()="conference publication"]]/mods:originInfo/mods:place/mods:placeTerm
//mods:mods[1][mods:genre[@authority="local"][text()="conferencePaper"]]/mods:originInfo/mods:place/mods:placeTerm
```

And the final XPath isn't included because it wouldn't be valid MODS:
`//mods:mods[1]/mods:relatedItem[@type="series"]/mods:name`

The only thing that could live here in a valid file would be the typical stuff that lives in `<mods:name>` at the top level, such as `<mods:namePart>`, `<mods:affiliation>`, etc. We might have to adjust that XPath because its possible that it was never correct to begin with.

d-r-p:
@bryjbrown I do not know if it helps, but:
> However some of the XPaths you listed I just can't figure out so I'm not sure how to add an element that satisfies them, such as:
    ```
//mods:mods[1][mods:genre[@authority="local"]="conferencePaper"]/mods:relatedItem/mods:titleInfo/mods:title
    ```
    ```
//mods:mods[1][mods:genre[@authority="marcgt"][text()="conference publication"]]/mods:originInfo/mods:place/mods:placeTerm
    ```
    ```
//mods:mods[1][mods:genre[@authority="local"][text()="conferencePaper"]]/mods:originInfo/mods:place/mods:placeTerm
    ```

I think you will need to create two different MODS files, each with a different `<genre>`. Instead of 
```xml
    <genre authority="rdacontent">text</genre>
```
write
```xml
    <genre authority="local">conferencePaper</genre>
```
respectively
```xml
    <genre authority="marcgt">conference publication</genre>
```
Then, the above XPATHs should retrieve:
1. Web of Knowledge (for the first `<genre>`)
2. bcc (for the second `<genre>`)
3. bcc (for the first `<genre>`)

DonRichards:
@bryjbrown  context for those values

Citeproc uses this as the "Event"
It returns the value for [`mods:genre[@authority="local"]="conferencePaper"]/mods:relatedItem/mods:titleInfo/mods:title`](https://github.com/Islandora/islandora_scholar/blob/7.x/modules/citeproc/includes/converter.inc#L240)
Unless '**islandora_scholar_xpaths_event_property**' value exist 

Citeproc uses this as the "Event **place**"
[`//mods:mods[1][//mods:mods[1][mods:genre[@authority="local"][text()="conferencePaper"]]/mods:originInfo/mods:place/mods:placeTerm`](https://github.com/Islandora/islandora_scholar/blob/7.x/modules/citeproc/includes/converter.inc#L268)
Unless this value exist
[`//mods:mods[1][mods:genre[@authority="marcgt"][text()="conference publication"]]/mods:originInfo/mods:place/mods:placeTerm`](https://github.com/Islandora/islandora_scholar/blob/7.x/modules/citeproc/includes/converter.inc#L261)

These are references to the original values.

DonRichards:
@bryjbrown I can use the MODS file to test this PR but it does crash the object level viewing. Not sure why or how that's even possible. Should a MODS file be able to crash the UI? This looks like a bug that separately should be looked into.

bondjimbond:
@DonRichards @bryjbrown I started attempting to build a MODS file with those Xpaths, but there's some really bad MODS in there (for example, "mods:date" is NOT A THING).

> Should a MODS file be able to crash the UI? 

It will if it's invalid XML.

DonRichards:
@bryjbrown @bondjimbond @d-r-p Do we fix the values now or push this through and address the already incorrect value in its own PR? I only ask because it looks like this might stall this PR and it looks like this isn't adding to or complicating an existing issue. 

And does anyone have a valid/well formed MODS to test with that holds values for all of these entries?

bondjimbond:
@DonRichards I don't think we can fix the default values with this PR, since it's a different issue (and a bigger can of worms)... I think we get this functionality sorted out first, merge it, then make sure that when the defaults get fixed they also get fixed in this form.

Re the MODS file: I'll work on it as soon as I have some time, hopefully today.

bryjbrown:
@bondjimbond `<mods:date>` is a thing: https://www.loc.gov/standards/mods/userguide/part.html#date

@DonRichards I have no idea whats going on with the new MODS record I provided you, the original version was valid according to the MODS 3.7 XSD, and removing the `<mods:nameIdentifier>` part makes it compatible with earlier versions.  You are correct though that it is crashing the object page load, which happens reliably with invalid MODS (which should be a separate ticket), even though every tool in my toolbox is saying that this record is valid MODS. Shrug.

DonRichards:
@bryjbrown Well part of the issue might be that the MODS file says 3.5 at the top but changing it doesn't fix the issue.

DiegoPino:
@bryjbrown @DonRichards is the MODS issue happening because its a mods Collection instead of a single mods record? @DonRichards can you pass it through islandora´s mods cleanup xslt? That should get rid of that piece. It´s totally valid, but depending on what things you have enabled in your object landing page (like marcxml, et. al) it can get weird/buggy when the mods record is wrapped in a collection tag

DonRichards:
@DiegoPino the cleanup_thesis_form_ISLANDORA1350.xslt ?

bondjimbond:
@bryjbrown Right, sorry, should've specified: mods:date is a thing, but only as a child of mods:part (and not as a direct child of mods:relatedItem, as in the xpath given).

Attached my own attempt at hitting every xpath. I'm not getting any crashes when viewing this one. 
[MODS.Record.2.xml.txt](https://github.com/Islandora/islandora_scholar/files/2329314/MODS.Record.2.xml.txt)

Getting an interesting error, though... having added the `<genre authority="endnote">` bit, I get the following:

> Notice: Undefined index: conference paper in _get_endnote_type() (line 409 of /var/www/drupal/sites/all/modules/islandora_scholar/modules/citeproc/includes/converter.inc).

d-r-p:
@DonRichards @bryjbrown @bondjimbond I do not mind if we push this through before correcting the wrong values. However, I reckon we should make absolutely sure that the system behaves _identically_ to before this PR if no XPath is changed (independently of whether the admin form has been saved or not). This way, users who already circumvented the wrongly defined (and previously hard-coded) default values will not be caught by surprise. Also, I feel that fixing those values later will be easier this way... Or do you disagree?

@DiegoPino Good idea with cleaning the MODS. But regarding the wrapping `<modsCollection>`, would it not be possible to get rid of that manually?

bondjimbond:
Doing some testing with my document... Looks like there are still a few elements not in it, but it's a start.

The Created Date field (//mods:mods[1]/mods:originInfo/mods:dateCreated) still produces the subTitle instead of the dateCreated. I've verified that dateCreated is in the right place in the document. Any ideas what's happening there?

I would change the heading "Collection Title" to "Series Title" since that's what it's looking for.

"Event Title" (//mods:mods[1][mods:genre[@authority="local"]="conferencePaper"]/mods:relatedItem/mods:titleInfo/mods:title) doesn't seem to be working... could anyone confirm that I've put my content in the expected elements?

"PubMed ID" (//mods:mods[1]/mods:identifier[@type="accession"]) -- why is this heading PubMed ID? It should be "Accession Number". PubMed ID would be @type="pmid".

"Types" - not a very explicit header. i'm honestly not sure what it's really looking for, but it is one of the Genre options. Does anyone have any idea where the content of this element is actually displayed?

"MARC genre type (relative child): chapter titles in a book" - very hard to understand this. What it's actually displaying is the title of the containing object (whether that's a book, journal, whatever).

"Names" under the Citeproc section: output is kind of weird -- "Usborne, Wade author." Is this the desired output?

"Collection Editor" - should this not be Series Editor?

bryjbrown:
My personal $0.02 is pretty much what @d-r-p said: merge the PR once it can behave identically to the pre-PR version, with configurable XPaths being the only difference. Changing the default XPaths is indeed a separate fight (and more importantly, one we can fight one XPath at a time).

DonRichards:
@bondjimbond 
Would changing it to "Endnote authority genre types" work?
MARC genre type is more or less the same from what I can see also changing it to "MARC authority genre types" might be more helpful. 

As for the "Collection Editor": Citeproc labels it as `array('editor' => 'collection-editor',)`. Should it stay or are those just random variable names?

DonRichards:
I pushed them to see if they look right and make sense.

d-r-p:
@bondjimbond
> Getting an interesting error, though... having added the `<genre authority="endnote">` bit, I get the following:
> > Notice: Undefined index: conference paper in _get_endnote_type() (line 409 of /var/www/drupal/sites/all/modules/islandora_scholar/modules/citeproc/includes/converter.inc).

Indeed, `_get_endnote_type()` only recognises the following keys (see [`modules/citeproc/includes/converter.inc@L396ff`](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-dc3682c7ee8afb241d9961a4b93eb82eL396):
```php
  $type_map = array(
    'Book Section' => 'chapter',
    'Book' => 'book',
    'Edited Book' => 'book',
    'Journal Article' => 'article-journal',
    // XXX:  This is a custom one for FJM...  Might need a hook?
    'Working Paper' => 'chapter',
    'Thesis' => 'thesis',
  );
```
I guess, yes, we do need a hook, or, at least, should harmonise that across helper functions.

d-r-p:
@DonRichards
> As for the "Collection Editor": Citeproc labels it as `array('editor' => 'collection-editor',)`. Should it stay or are those just random variable names?

I think it should stay. The key in the array seems to be the role as specified in `<role><roleTerm>` (see [`convert_mods_to_citeproc_json_name_role()`](https://github.com/Islandora/islandora_scholar/blob/7.x/modules/citeproc/includes/converter.inc#L662-L698)), whereas the value specifies the [corresponding CSL variable](http://docs.citationstyles.org/en/1.0.1/specification.html#name-variables).

d-r-p:
<strike>
@bondjimbond @DonRichards

> The Created Date field (//mods:mods[1]/mods:originInfo/mods:dateCreated) still produces the subTitle instead of the dateCreated. I've verified that dateCreated is in the right place in the document. Any ideas what's happening there?
</strike>


I just observed that the original value for dateCreated used to be `//mods:originInfo/mods:dateCreated` instead of `//mods:mods[1]/mods:originInfo/mods:dateCreated`. Actually, the `'islandora_scholar_xpaths_created_date'` variable still defaults to the old value [here](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-04c6e90faac2675aa89e2176d2eec7d8R73), [here](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-17162048acce03c2e196eaca3b9e8d1bR108), [here](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-dc3682c7ee8afb241d9961a4b93eb82eR725) and [here](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-9a90d2a2ffb15eee5f6100961f96e1ebR78), but got changed only in the [admin form](https://github.com/Islandora/islandora_scholar/pull/299/files#diff-0cdd19abe14e51fe9e7f061ec72c86d3R621). <strike>I have doubts, though, that this is the underlying issue here...</strike>
<strike>
@bondjimbond Is this also what you see in Citeproc-redered citations, or just in the admin form?
[EDIT: @bondjimbond Is there a `<dateCreated>` in [your MODS](https://github.com/Islandora/islandora_scholar/files/2329314/MODS.Record.2.xml.txt)? I cannot seem to find it... Only `<dateIssued>`, `<dateCaptured>`, `<dateOther>` and `<date>`.]
</strike>
[EDIT2: Scratch the part about the issue - it looks like @DonRichards [found and corrected](https://github.com/Islandora/islandora_scholar/pull/299/commits/41103b43ed3e8b987535acc5dc3e60167a81579e) the problem. However, the different default values for `'islandora_scholar_xpaths_created_date'` might still be problematic.]

d-r-p:
@bondjimbond 
> "Event Title" (`//mods:mods[1][mods:genre[@authority="local"]="conferencePaper"]/mods:relatedItem/mods:titleInfo/mods:title`) doesn't seem to be working... could anyone confirm that I've put my content in the expected elements?

I think your `<genre>` tags should not reside inside `<relatedItem>`, but should be children of `<mods>`. Moreover, I observed that you have built a MODS with several `<relatedItem>` tags - not sure if the resulting value arrays will be handled well. May I suggest to create different MODS for different document types?

> "PubMed ID" (`//mods:mods[1]/mods:identifier[@type="accession"]`) -- why is this heading PubMed ID? It should be "Accession Number". PubMed ID would be @type="pmid".

This is [legacy code](https://github.com/Islandora/islandora_scholar/search?q=accession&unscoped_q=accession) which we should fix when we attack the default values one by one, no?

DonRichards:
That last commit should have probably have been it's own PR. This gives feedback to the user for when _get_endnote_type() can't process the endnote type. @bryjbrown  @d-r-p @DiegoPino @bondjimbond Do you guys have any further changes left for this PR? Have I missed anything?

DonRichards:
### Recap with Questions,
we know some of the XPaths are not correct and should probably be corrected in future PRs. All of the values in this [MODS Record.xml](https://github.com/Islandora/islandora_scholar/files/2332718/MODS.Record.3.replace.xml.txt) appear to load values correcrtly into the admin config page '/admin/islandora/solution_pack_config/scholar/xpaths'.

### Questions for everyone
- [ ] Is the naming of each entry intuitive / useful?
- [ ] Does this have any effect on existing items? 
- [ ] Is the grouping of the entries intuitive / useful?
- [ ] Is there any entry descriptions that should be changed?
- [ ] Is there any garbage code that shouldn't be left in like a missed dsm() message or random code comment?
- [ ] Does this have any negative effect on ingesting items?
- [ ] Is there anything that should be modified before moving forwards with this?

bryjbrown:
@DonRichards Just chiming in to let you know I haven't forgotten about this, I've been picking through it and I'm trying to arrange my thoughts about how the "XPATH CONFIGURATION" admin page should be arranged, I just need a bit more time to sculpt my feelings into something more actionable. 

I didn't find any leftover watchdogs/dsms though, but I'll give it another look.

DonRichards:
@bryjbrown If it helps, we can always rearrange them. The way I assigned them is by adding them to 1 of 4 groups. 

All it takes to move one from Google Xpaths to Coins Xpaths is replace the word google with coins in "islandora_scholar_xpaths_google" part of the $form[ID] and the ajax callbacks as well. No need to move code. It will show up as expected at that point.
```php
$form['islandora_scholar_xpaths_google'] += _islandora_scholar_add_variable('islandora_scholar_xpaths_genre', '//mods:mods[1]/mods:genre', array(
    '#type' => 'textfield',
    '#title' => t('Genre'),
    '#description' => t('Default Value: //mods:mods[1]/mods:genre <br/>XPath to use for genre.'),
    '#states' => array(
      'visible' => array(
        ':input[name="islandora_scholar_xpaths_enabled"]' => array('checked' => TRUE),
      ),
    ),
    '#ajax' => array(
      'callback' => 'islandora_scholar_xpaths_google_callback',
      'wrapper' => 'islandora_scholar_xpaths_google_title',
    ),
   '#field_suffix' => (!empty($form_state['values']['islandora_scholar_xpaths_demo']) ? _mods_id('islandora_scholar_xpaths_genre', $form_state['values']['islandora_scholar_xpaths_genre'], check_plain($form_state['values']['islandora_scholar_xpaths_demo'])) : t("Modify to refresh value (PID for demo object is required).")),
  ));
```
So moving is very simple. Aside from rearranging, is there anything else that isn't functioning as expected? Should there be any other changes made?

DonRichards:
@bryjbrown So how are we doing on this? Should we aim to have this merged in before code freeze?

bondjimbond:
> So how are we doing on this? Should we aim to have this merged in before code freeze?

Yes, please. I haven't found time to test the latest iteration, but I'll give it a shot this week. I would love for this beast to be slayed before code freeze.

DonRichards:
Thanks @bondjimbond, me too.

bryjbrown:
@DonRichards Okay, finally found some time to finish reviewing it. Here are my recommendations:

1. At the top of the XPath menu tab, there's some text that has links to additional documentation. While I think these are helpful, I think they should be moved out of the admin page and into the README.
screenshot from 2018-09-12 16-59-16

2. In the README, there is a list of all configurable XPaths, and they all have nested lists with Default Value and Context, but Context is blank and doesn't have anything in it. Not sure what this is for, but Context should either have a value, or be removed. Additionally, I think this would be a good place to add any information about how and where that XPath gets used if we happen to have that info.

3. Each section on the XPath configuration page has a short description, like "XPath for Coins submodule to construct the first entry in an array of subtitle entries." or "XPath to use for the created_date." Most of these are either confusing or unhelpful, and they clutter up the space on the page. I'd be in favor of moving these descriptions to the README instead.

Other than these nitpicks, I think this PR is solid and should be ready for this release :+1:

DonRichards:
@bryjbrown thanks! 2 and 3 are right on the money for what "context" should be. I just didn't completely know what to write. I'll make those adjustments.

DonRichards:
Just committed some of those. A couple of the links your suggested to move are relative to the menu so I thought I'd leave them until it was discussed. Take a look and let me know if more text should be dropped.

DonRichards:
Looks like it's ready to review. Is this what you were suggesting @bryjbrown ?

bryjbrown:
@DonRichards Awesome, I'm actually out of town right now but I will give this a test Thursday when I'm back :+1:

includes/admin.form.inc
```diff
@@ -351,22 +353,1290 @@ function islandora_scholar_admin_pdf_form($form, &$form_state) {
...
+    $islandora_scholar_xpaths_google_description = t('These XPaths are shared within most of the submodules. To see the MODS value after editing an entry, tab to the next field.');
+    $islandora_scholar_xpaths_coins_description = t('These XPaths are shared within most of the submodules. To see the MODS value after editing an entry, tab to the next field.');
+    $islandora_scholar_xpaths_citeproc_description = t('These XPaths are shared within most of the submodules. To see the MODS value after editing an entry, tab to the next field.');
+  }
+  $form = array(
+    'islandora_scholar_xpaths' => array(
+      '#type' => 'fieldset',
+      '#title' => t('Islandora Scholar Xpath configuration'),
+      '#description' => t("Set custom XPaths for Islandora Scholar. <br/><br/>
+      Additional XPath Configurations !gs_search_xpaths_configuration & !library_catalog_search_xpaths_configuration<br/><br/>",
```
bryjbrown: I'm still not a fan of this line existing on the config page. I think this info belongs in the README only.

bryjbrown:
@DonRichards Finally got to take a look at this, it all looks good except for one line, which I made a review for.

includes/admin.form.inc
bryjbrown: I hate GitHub's reviews UI, so in case you can't tell what I'm pointing to, its the part of the admin page UI that says `Additional XPath Configurations Google Scholar Search XPaths & Library Catalog Search XPaths Configuration Page`.

IMHO thats taking up space on an already busy page with information that rightly belongs in the README or JIRA documentation.

DonRichards: OK, no problem. I can make that move. I left it in because of the name "XPaths" implies all XPaths are located on that tab but there are a few elsewhere.

bondjimbond:
I just noticed something odd... An error message when viewing an object that does not appear as a normal error message:

> Endnote genre needs to be limited to the mapped array(Book Section, Book, Edited Book, Journal Article, Working Paper, or Thesis). Please contact the system administrator about this message.

Was this always there? (I had set the object's genre to "conference paper".) In the 7.x branch I get a different error:

> Notice: Undefined index: conference paper in _get_endnote_type() (line 409 of /var/www/drupal/sites/all/modules/islandora_scholar/modules/citeproc/includes/converter.inc).

So:

1. Was this error message introduced in this branch?
2. If so, can we make it a normal error message so that it can be suppressed from public view by choosing the "suppress errors/warnings" option in Drupal?

bondjimbond:
Normal error message:
screen shot 2018-09-24 at 3 25 31 pm

Error message on this branch:
screen shot 2018-09-24 at 3 25 41 pm

DonRichards:
@bondjimbond this looks like it is expected behavior prior to this PR and is on line 413.
```php
...
echo '<br/>' . t('Endnote genre needs to be limited to the mapped array(Book Section, Book, Edited Book, Journal Article, Working Paper, or Thesis). Please contact the system administrator about this message.') . '<br/><br/>';
...
```
To keep the scope of this on track I'd suggest not changing within this PR if it isn't completely necessary.

bryjbrown:
@DonRichards If it was expected behavior prior to this branch, why is @bondjimbond getting a different one? @bondjimbond, can you double check that you are using the latest 7.x branch to compare to?

modules/citeproc/includes/converter.inc
```diff
@@ -402,11 +402,16 @@ function _get_endnote_type(SimpleXMLElement $mods) {
...
+  $types = $mods->xpath(variable_get('islandora_scholar_xpaths_types', '//mods:mods[1]/mods:genre[@authority="endnote"]'));
 
   if (!empty($types)) {
     $type = $types[0];
-    $output = $type_map[(string) $type];
+    if (isset($type_map[(string) $type])) {
+      $output = $type_map[(string) $type];
+    }
+    else {
+      echo '<br/>' . t('Endnote genre needs to be limited to the mapped array(Book Section, Book, Edited Book, Journal Article, Working Paper, or Thesis). Please contact the system administrator about this message.') . '<br/><br/>';
```
bryjbrown: So this WAS introduced this PR, so it will have to be changed. Raw HTML messages posted above the rendered portion of the page are okay for debugging, but we can't release something like this. We either need to add a default type of some sort (my vote if we can figure out what that type would be, "generic" may be an option) or at the very least send the error to watchdog or something instead of presenting it user-side.

bryjbrown (changes requested):
We need to fix that line, even if the underlying error precedes this PR, the new fix for it doesn't and we need to change that.

modules/citeproc/includes/converter.inc
DonRichards: @bryjbrown @bondjimbond  MAN. Guys, I'm sorry! I completely didn't remember adding this. I just looked it up, I see that I did but seriously don't even remember why it's here. I'll be spending the next few minutes reviewing it and make the necessary changes.

DonRichards:
Well that was embarrassing.

DonRichards:
@bryjbrown Does that work?

bryjbrown:
@DonRichards This is better, but to be honest I'm still not crazy about presenting user-facing error messages for things that aren't REALLY an error, just a mismatch between local practice and what Scholar expects (other possible Endnote types).

Before this PR, the code expected that either there isn't a `<genre authority="endnote">` or there is, and the value corresponds to one of the keys in the `$type_map` array. As @bondjimbond points out, this resulted in an automated `Notice: Undefined index` error message, which is the worst case scenario.

This PR moves in the right direction by making sure that the value in `<genre authority="endnote">` corresponds to a key in the `$type_map` array, and throws a `drupal_set_message` error if it doesn't. This is certainly better than a standard error message from Drupal, but I still disagree that its an actual error and strongly disagree with presenting this error to a user viewing the object.

My suggestions are:
1. We log this as a watchdog warning so that admins can pick up on this issue in the logs
2. Setting the value to `generic` ([gleaned from the list of types here](http://libguides.utoledo.edu/c.php?g=284330&p=1895352)) if the value doesn't match anything in the `$type_map`
3. We just don't do anything if the `<genre authority="endnote">` doesn't match anything in the `$type_map`. We simply pretend there is no `<genre authority="endnote">` rather than present an error or try to make a guess.

I'm fine with either option. @Islandora/7-x-1-x-committers let me know what you think is the best way to handle this issue.

bondjimbond:
> Endnote genre needs to be limited to the mapped array(Book Section, Book, Edited Book, Journal Article, Working Paper, or Thesis)

After this, the next thing we need is an admin menu to map preferred genre terms to these, so that sites can use standard genre terms (marcgt, tgm2, etc.) without worrying about how Citeproc interprets them.

bryjbrown:
@bondjimbond Can you test this latest commit with the object that triggered the warning for your previously?

bondjimbond:
@bryjbrown The notice now gets pushed to the log, which is good, but the message gets cut off.

> Endnote genre needs to be limited to the mapped array(Book Secti

The above appears under "Type", when it should really go under "Message" (and not get cut off). "Type" could be something like "Endnote genre not recognized" or something like that.

screen shot 2018-09-26 at 11 06 54 am

DiegoPino:
@bryjbrown @bondjimbond  "Type" tends to be in most of Drupal 7  modules a more general concept, like "Islandora Scholar", etc. Mostly because you can filter by type and the severity is already an option/flag. Message should be the current "type". i feel you are invoking the watchdog with arguments in the wrong order.

bryjbrown:
@bondjimbond You're right, this watchdog is formatted wrong. Should be:
`watchdog('Islandora Scholar', 'Endnote genre needs to be limited to the mapped array(Book Section, Book, Edited Book, Journal Article, Working Paper, or Thesis).');`

DonRichards:
@bryjbrown I just realized your suggestion didn't include a severity phrase. Should it not include the severity part "WATCHDOG_WARNING"?

bryjbrown:
@DonRichards It could, watchdog defaults to "notice" which I think is fine, but if you want to make it a warning you'll need to specify that as the fourth argument and pass in an empty array as the third argument (for extra variables). Having the WATCHDOG_WARNING contant as the third argument gets it out of order and causes the `/admin/reports/dblog` page to crash when you try to view the messages. 

So to review, we need either one of the following (which I have tested):
- Notice (default): `watchdog('Islandora Scholar', 'Endnote genre needs to be limited to the mapped array(Book Section, Book, Edited Book, Journal Article, Working Paper, or Thesis).');`
- Warning: `watchdog('Islandora Scholar', 'Endnote genre needs to be limited to the mapped array(Book Section, Book, Edited Book, Journal Article, Working Paper, or Thesis).', array(), WATCHDOG_WARNING);`

See https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/watchdog/7.x for more info about the specific order of watchdog arguments.

DonRichards:
@bryjbrown How are you testing this? I'm not sure I'm doing the same steps or something. I've not seen that page crash yet. I see the 500 errors caused but an invalid input but no crashing for me.

bryjbrown:
@DonRichards I'm testing like this:

0. Make sure you are starting your testing on the main 7.x branch
1. Create an object for testing
2. Download the MODS datastream
3. Edit it locally to have someting like `<genre authority="endnote">This is not a valid genre</genre>`
4. Replace the object's MODS datastream with the edited copy
5. Refresh the page for the object's standard view. You should get a `Notice: Undefined Error` message
6. Switch branches to this current branch, and change the watchdog line in question to one of the options from https://github.com/Islandora/islandora_scholar/pull/299#issuecomment-424801020, then `drush cc all`
7. Refresh the page for the object view
8 . Open up the `/admin/reports/dblog` recent log messages reports view, preferably in a new tab. See that the appropriate watchdog message was logged
9. Discard these changes and check out the original `converter.inc` file that represents the current state of this PR, and `drush cc all`, then refresh the object page.
10. Check the `/admin/reports/dblog` again, and you'll get this error message:
screenshot from 2018-09-26 14-09-32

Once that malformed watchdog error hits the log, theres nothing I can do to make the logs accessible again save for destroying and rebuilding the VM. Either of the options from https://github.com/Islandora/islandora_scholar/pull/299#issuecomment-424801020 should correctly log the error and keep `/admin/reports/dblog` accessible.

bondjimbond:
My 2c goes to "Warning"

bryjbrown:
@bondjimbond Me too. I suppose that just means that the existing watchdog needs an `array(), ` as the third argument, between the message and the `WATCHDOG_WARNING`.
```
watchdog('Islandora Scholar', 'Endnote genre needs to be limited to the mapped array(Book Section, Book, Edited Book, Journal Article, Working Paper, or Thesis).', array(), WATCHDOG_WARNING);
```

DonRichards:
So does it look good now or are you still seeing an error? 
```php
watchdog('Islandora Scholar', 'Endnote genre needs to be limited to the mapped array(Book Section, Book, Edited Book, Journal Article, Working Paper, or Thesis).', array(), WATCHDOG_WARNING);
```

bondjimbond:
Looks good to me!

bryjbrown:
@DonRichards :+1: I formally dismiss my review. I'm okay to merge this now. Great work y'all! I'll merge in 24 hours.

bryjbrown (approved):

DonRichards:
Thanks everyone! This was a big one.

DonRichards:
@bryjbrown *nudge :-D

PR #302, opened 2018-09-28
ISLANDORA-2027: Implements Gail Lewis fix for namespace filtering on scholar embargoes

bryjbrown:
**JIRA Ticket**: https://jira.duraspace.org/browse/ISLANDORA-2027

Copy/pastes @glewisorg's solution to applying namespace filtering to the SPARQL queries asking for all embargoed objects

# What's new?
Adds code from https://github.com/FLVC/islandora_scholar/tree/flvc to the core 7.x branch so that everyone can filter out embargoed items that are outside of a multisite's allowed namespaces the same way FL-Islandora can.

Use case: you are running a multisite where individual sites are minting new PIDs in their own namespace, and they are all using Islandora Scholar Embargo. Currently, the SPARQL query that retrieves all embargoed items doesn't filter out restricted namespaces, so site A will see site B's embargoed items under `/admin/islandora/tools/embargo/list`, even though site B's namespace is not listed in site A's allowed namespaces (`/admin/islandora/configure`).

With this PR, the SPARQL query will filter out all the embargoed items outside of a site's allowed namespace.

# How should this be tested?
1. Spin up a fresh VM and set the Islandora Scholar module's branch to 7.x
2. Go to a folder and create a new object in the default namespace (for example, "nsa")
3. Go to the collection management tab for the parent collection, and set a brand new namespace to be the default for that same object type (for example, "nsb")
4. Create a new object of the same type so that it has the new namespace
5. Apply embargoes to both objects
6. Go to `/admin/islandora/tools/embargo/list` and verify that both objects are visible
7. Go to `/admin/islandora/configure` and click the "Namespaces", and check "Enforce namespace restrictions", and add make sure the first object's namespace is on the list of allowed namespaces, but the second object's namespace is not
8. Go back to the original collection that both objects are in, and verify that the second object no longer appears (Islandora is filtering it out)
9. Go back to `/admin/islandora/tools/embargo/list` and verify that, despite Islandora filtering out the second object's namespace, the second object is still appearing in the list of embargoed objects
10. Switch the Islandora Scholar module's branch to this PR, and run `drush cc all`
11. Refresh `/admin/islandora/tools/embargo/list` and verify that the second object with the disallowed namespace no longer appears in the list of embargoed items.

# Additional Notes:
Many thanks to @glewisorg for the solution. I am but a humble copy/paster of her original efforts.

# Interested parties
@Islandora/7-x-1-x-committers @bondjimbond @DonRichards @DiegoPino @rosiel

### DIFF
--- a/modules/islandora_scholar_embargo/includes/embargo.inc
+++ b/modules/islandora_scholar_embargo/includes/embargo.inc
@@ -460,6 +460,7 @@ SELECT ?obj ?date
 FROM <#ri>
 WHERE {
   ?obj is:embargo-until ?date
+  !namespace_filter
 }
 EOQ;
   if ($count) {
@@ -467,6 +468,10 @@ EOQ;
 ORDER BY ?obj
 EOQ;
   }
+  $namespace_filter = islandora_scholar_embargo_get_namespace_filters();
+  $current_embargo = format_string($current_embargo, array(
+    '!namespace_filter' => !empty($namespace_filter) ? $namespace_filter : '',
+  ));
   $connection = islandora_get_tuque_connection();
   $ri = $connection->repository->ri;
   return $count ? $ri->countQuery($current_embargo, 'sparql') : $ri->sparqlQuery($current_embargo, 'unlimited', '0');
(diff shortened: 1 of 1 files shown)


DonRichards:
During the committers' call @DiegoPino said he'd be interested in looking at this.

bryjbrown:
@DonRichards I see you added the "enhancement" lable, but this is a bug according to the JIRA ticket.

DiegoPino:
Yep, can do tomorrow afternoon, for sure!

DonRichards (approved):
Following those exact instructions I get the expected results.

bondjimbond:
@DonRichards It's been more than 24 hours since you approved... Ready to merge?

bondjimbond:
Well, I've tested as well, and approve... if I don't hear from @DonRichards today I'll hit that merge button.

bondjimbond (approved):

bondjimbond:
@bryjbrown Does this PR address namespace-related issues outside that specific management screen?

For example, here's what I've been getting when changing one site's permitted roles for viewing embargoed objects:

screen shot 2018-10-17 at 10 48 59 am

I'm in the rdc site, which does not have access to objects in the tru namespace. In fact, I'm not sure that the rdc site has any embargoed items at all, but it is still updating the embargoed items at all of my other sites.

I also get emails:
screen shot 2018-10-17 at 10 51 35 am

Note the above examples are happening in the live site on the 7.x branch, not your fix branch... I'd like to know if this is consciously a part of the fix before I go through the effort of testing. :)

bryjbrown:
@bondjimbond I believe the batch process that re-applies the embargoes should respect the namespace filter in my PR branch since the batch process uses the same core function this PR improves to gather the list of all embargoed items to work on. Definitely test this and let me know how it goes.

As for the embargo expiry notifications, the SPARQL query that gathers up the expiring embargoes is separate from the namespace filtered one, but I don't think that matters because that happens automatically on the back end through cron and not through exposed user interactions.

Also, the emails you are getting might be because of a specific rule that you have set up or because you are the object's "owner" (the one who ingested it). The owner gets an embargo expiry notification unless you have told it otherwise.

bondjimbond:
@bryjbrown Cool, it works. Applies changes only to the objects with approved namespaces. This is good to go.

DonRichards:
@bondjimbond Sorry I didn't respond. I missed the notifications.

PR #316, opened 2019-03-08
Fix Travis errors

bondjimbond:
**JIRA Ticket**: (https://jira.duraspace.org/browse/ISLANDORA-2390)

Reorders commands in the .travis.yml file and adds "entity" dependency to Islandora Scholar Embargo .info file.

This resolves the problem of Travis failing to build, because the "rules" module (a dependency of some Scholar submodules) requires "entity_token", which is a submodule of "entity", before "entity" itself is downloaded.

# What's new?

Just a change in order, plus making the "entity" dependency of Islandora Scholar Embargo explicit (rather than an invisible dependency-of-a-dependency).

Should not impact anything.

# How should this be tested?

See if Travis builds successfully on this PR.

# Interested parties
@Islandora/7-x-1-x-committers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,7 @@ before_install:
   - ln -s $HOME/islandora_bookmark sites/all/modules/islandora_bookmark
   - ln -s $HOME/islandora_batch sites/all/modules/islandora_batch
   - drush cc all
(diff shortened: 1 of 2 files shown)


bryjbrown (approved):

bryjbrown:
@bondjimbond I'm not sure why switching the order that things get enabled in would affect this, but I can't argue with the results. This is helping Travis to pass when the other PRs continue to fail, so I'm in favor of merging this ASAP. Any objections?

bondjimbond:
@bryjbrown Looks like no objections! Once this is merged, I think there are a few other modules that are hitting similar roadblocks with Travis.

PR #318, opened 2019-04-02
ISLANDORA-2409: Selectable colorspace for PREVIEW/TN

BrandonMorr:
**JIRA Ticket**: https://jira.duraspace.org/browse/ISLANDORA-2409

Adds a dropdown to the solution pack config UI which allows users to select from a list of colorspaces to use when creating TN/PREVIEW derivatives. The intention of this pull is to allow users to choose a more reliable/consistent colorspace profile for PDF derivatives. The currently hardcoded RGB colorspace appears to create substantially darker derivatives compared to their source file.

# What's new?

* Adds dropdown select to configure colorspace configuration for PREVIEW & TN derivative datastreams
* Profiles include RBG, sRGB, CYMK & Grayscale
* Update hook preserves existing colorspace for older systems

# How should this be tested?

Prior to pulling down the changes, ingest a test PDF (preferably a PDF with light colors on the 1st page) and observe that the PREVIEW and TN datastreams colors are far darker than the source PDF's 1st page's colors.

Pull down the changes and go to the PDF scholar solution pack config UI. Confirm the sRGB colorspace is set for both PREVIEW and TN and submit the form. Ingest another test PDF and observe the PREVIEW and TN datastreams and confirm that they are closer to the original source PDF.

# Additional Notes:

Similar pull request for PDF sol pack: https://github.com/Islandora/islandora_solution_pack_pdf/pull/138

### DIFF
--- a/includes/derivatives.inc
+++ b/includes/derivatives.inc
@@ -47,7 +47,7 @@ function islandora_scholar_add_tn_derivative(AbstractObject $object, $force = FA
     $height = variable_get('islandora_scholar_thumbnail_height', 200);
     $colorspace = variable_get('islandora_scholar_thumbnail_colorspace', 'sRGB');
     $file_uri = islandora_scholar_get_derivative_source_as_unmanaged_file($object, $hook['source_dsid']);
-    $results = islandora_scholar_add_jpg_derivative($object, $file_uri, 'TN', $width, $height, $force, $hook['source_dsid']);
+    $results = islandora_scholar_add_jpg_derivative($object, $file_uri, 'TN', $width, $height, $force, $hook['source_dsid'], $colorspace);
     file_unmanaged_delete($file_uri);
     return $results;
   }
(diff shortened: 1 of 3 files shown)


qadan (approved):
Acceptably improves from the usage of Adobe RGB colorspace 👍

jordandukart:
Bump @Islandora/7-x-1-x-committers and @dannylamb

includes/derivatives.inc
```diff
@@ -193,12 +197,12 @@ function islandora_scholar_add_preview_derivative(AbstractObject $object, $force
  *
  * @see hook_islandora_derivative()
  */
-function islandora_scholar_add_jpg_derivative(AbstractObject $object, $file_uri, $dsid, $width, $height, $force, $source_dsid) {
+function islandora_scholar_add_jpg_derivative(AbstractObject $object, $file_uri, $dsid, $width, $height, $colorspace, $force, $source_dsid) {
```
jonathangreen: Should we be changing the function signature here in case someone is calling this in their own module? 

```php
function islandora_scholar_add_jpg_derivative(AbstractObject $object, $file_uri, $dsid, $width, $height, $force, $source_dsid, $colorspace = NULL)
```

Might be more appropriate?

I don't see any uses in core code, or my own code, so I'm not super concerned, but I thought it was worth mentioning.

jonathangreen (commented):
This looks fine to me, just one comment about a changed function signature.

includes/derivatives.inc
BrandonMorr: Definitely agree! Also caught some missing stuff. Thanks for the feedback 👍

bryjbrown: +1 to @jonathangreen's suggestion.

jonathangreen:
LGTM pending travis. 👍🏻

jonathangreen (approved):
Start merge ⏱

BrandonMorr:
@jonathangreen merge time? :)

PR #325, opened 2019-06-28
Enable Islandora Google Scholar on Scholar install and update

bondjimbond:
**JIRA Ticket**: (https://jira.duraspace.org/browse/ISLANDORA-2448)

On installation or update of Islandora Scholar, checks if Islandora Google Scholar is enabled. If not, it does so. (And lets the user know that this happened.)

# What's new?

As described above. The Islandora IR Interest Group expressed a desire to default to turning on Islandora Google Scholar, rather than relying on people reading the documentation and realizing that they should turn on the module. Not having it turned on means Google Scholar will probably not index this content.

# How should this be tested?

- If Islandora Google Scholar is enabled, disable it
- Pull this PR and run update.php
- See message, confirm that Islandora Google Scholar now enabled.
- Disable Islandora Scholar and everything that relates to it.
- Enable it again.
- Check whether Islandora Google Scholar is enabled.

# Interested parties
@Islandora/7-x-1-x-committers @bryjbrown

### DIFF
--- a/islandora_scholar.install
+++ b/islandora_scholar.install
@@ -13,6 +13,9 @@
 function islandora_scholar_install() {
   module_load_include('inc', 'islandora', 'includes/solution_packs');
   islandora_install_solution_pack('islandora_scholar');
+  $modules = array('islandora_google_scholar');
+  $enable_dependencies = TRUE;
+  module_enable($modules, $enable_dependencies);
 }
 
 /**
@@ -148,4 +151,16 @@ function islandora_scholar_update_7103(&$sandbox) {
   variable_set('islandora_scholar_thumbnail_colorspace', 'RGB');
   $t = get_t();
   return $t('Set colorspace configuration to RBG to maintain existing profile.');
+}
+
+/**
+ * Ensure Islandora Google Scholar is enabled.
+ */
+function islandora_scholar_update_7104(&$sandbox) {
+  if (!module_exists('islandora_google_scholar')) {
+    $modules = array('islandora_google_scholar');
+    $enable_dependencies = TRUE;
+    module_enable($modules, $enable_dependencies);
+    return t('The Islandora Google Scholar module has been enabled by default. If you wish to disable it you may do so at admin/modules.');
+  }
 }


bryjbrown (dismissed):

bryjbrown:
@bondjimbond Despite what Anurag said in the call, I have some serious reservations about enabling modules automatically like this when they aren't a true dependency and its actually more of a suggestion. I feel like it takes away control from the user, and there may be users who purposely decided not to run Islandora Google Scholar that would be frustrated with this update automatically enabling it. However, that seems like a much smaller chunk of folks that the ones who don't even know they aren't running it now or in the future, so I'm okay with this specific instance of automatically enabling a non-required module as part of an update. Does anyone else have any feelings about this? @dannylamb @DiegoPino @DonRichards @qadan?

I approved the change, code looks well written, but I didn't test it and I'm afraid I don't have the time to this week.

bondjimbond:
No negative comments on the idea at the Coordinating Committee meeting.

DonRichards:
I agree with @bryjbrown comments, I'm still not clear how enable by default would apply but I have a concern with this PR. When you pull in the git commit does it remove the submodule or does it leave the original files in place? This could be a concern, I wasn't sure but thought the update function should check. Just a question/thought.

bondjimbond:
@DonRichards I'm not sure what you mean...

> When you pull in the git commit does it remove the submodule or does it leave the original files in place? 

No files are removed. All this does is check whether or not the submodule is enabled, and enables it if not.

bryjbrown:
@DonRichards This PR doesn't remove any files, Islandora Scholar should have Islandora Google Scholar as a submodule as it always has. Currently, when you enable Islandora Scholar, Islandora Google Scholar isn't automatically enabled. This PR is trying to make it so that it is, but thats the only change.

From a code standpoint this looks good, but for some reason I cannot get it to work when testing. I've tried disabling the module and re-enabling it, building the VM without Scholar enabled and then enabling it (Scholar has circular dependencies that make it impossible to uninstall once enabled), and running the update.php script with Islandora Google Scholar disabled. All of these actions should have triggered Islandora Google Scholar to be enabled, but each time it was not. I'm not sure why, because when I run the code you have in hook_install() manually with `drush php-eval` it DOES enable Islandora Google Scholar. Weird! @bondjimbond, you tested all this on your own system and it works for you, right? @DonRichards any chance you could fire this up and let us know what it does for you?

Furthermore, I don't think that hook_install() runs when you enable a module that was already installed and then disabled, it only runs when you are enabling a module for the first time, see https://api.drupal.org/api/drupal/modules%21system%21system.api.php/function/hook_install/7.x:
> The hook will only be called the first time a module is enabled or after it is re-enabled after being uninstalled.

Even furthermore than that, its a bit misleading to install a module automatically that isn't listed as a dependency, because the user has no indication, either via UI or `drush pm-enable`, that Islandora Google Scholar will automatically be enabled.

I'm not sure if these last two are dealbreakers or not, but I only see one option that solves all the problems: list Islandora Google Scholar as a dependency of Islandora Scholar. This would make sure it is ALWAYS on, though its subjective whether that would be a bug or a feature. Its not LITERALLY a dependency, but we are kind of treating it like it is with this PR anyway by enabling it automatically without input from the user. If we're going to enforce our opinions, why not make them an offer they can't refuse? Requesting input from @dannylamb @DiegoPino and @whikloj for opinions on this.

bondjimbond:
> I only see one option that solves all the problems: list Islandora Google Scholar as a dependency of Islandora Scholar. This would make sure it is ALWAYS on, though its subjective whether that would be a bug or a feature.

I would be totally happy with that choice. While I often hear that people don't want more dependencies, Islandora Google Scholar does nothing substantial beyond adding metatags... Is there any reason someone using Scholar would explicitly NOT want those tags?

jonathangreen:
Instead of automatically enabling this module, if it's going to be a requirement to have this module installed would it make sense to roll the module into the main scholar module instead of having an implicit dependancy that is enabled by the scholar module? 

When thinking about these things, I'm a big fan of the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). To me its not an expected behaviour to have a module enable another module be default, but rolling the modules together makes sense and seems like a less surprising thing to do for the site administrator.

bryjbrown:
I think @jonathangreen's suggestion here is honestly the right thing to do in the grand scheme of things, and its not like Islandora Google Scholar is a big module. Its basically a .info file, a .install file that sets up some new variables, and a sub-300 line .module file. This could (and probably should) be folded into the main module.

My only concern with this route would be what happens when someone updates to a new version of the Scholar module where all the code is playing musical chairs, but I assume any of those issues could be solved via an update hook?

bondjimbond:
@bryjbrown Sounds right to me.

If someone else wants to take this on I'd appreciate it, because I won't have a chance to do any more Islandora work for the next month.

bryjbrown:
@bondjimbond KABOOSH https://github.com/Islandora/islandora_scholar/pull/326

PR #326, opened 2019-07-12
Assimilate Islandora Google Scholar in to Islandora Scholar

bryjbrown:
**JIRA Ticket**: https://jira.duraspace.org/browse/ISLANDORA-2448

Moves all Islandora Google Scholar functionality into Islandora Scholar module, along with an update hook to disable the Islandora Google Scholar module.

# What's new?
Islandora Google Scholar's variables were already included in Islandora Scholar's .install file uninstall hook, so no work needed there.

Islandora Google Scholar's utility functions for building and embedding meta tags was moved into islandora_scholar/includes/google_scholar.inc.

Those functions got called in Islandora Google Scholar's CMODEL_PID_view_object() hooks, these are now in the same place but in islandora_scholar.module's CMODEL_PID_view_object() hooks.

# How should this be tested?
Boot up a VM with the current 7.x HEAD Scholar module, then pull this PR. You should immediately see error messages about the Islandora Google Scholar module being missing. Run /update.php, and this error message should go away. View an object with robust metadata (insufficient metadata will stop the meta tags from displaying, its always been this way) and verify that the meta tags are in the HTML source.

# Additional Notes:
This PR supercedes https://github.com/Islandora/islandora_scholar/pull/325

# Interested parties
@Islandora/7-x-1-x-committers @bondjimbond @DonRichards @jonathangreen


bryjbrown:
Whoops I forget to remove references to islandora_google_scholar from .travis.yml, so all the tests failed. Fixed now.

# RESUME ASSIMILATION PROTOCOLS
robopicard

jonathangreen:
I *think* (but I'm not sure) as well as disabling the submodule needs to be uninstalled with [`drupal_uninstall_modules`](https://api.drupal.org/api/drupal/includes%21install.inc/function/drupal_uninstall_modules/7.x).

bryjbrown:
@jonathangreen I think you are correct, sir. Fixing now.

bryjbrown:
@jonathangreen This may not be what we want to do, since uninstalling the module would run the hook_uninstall() functions and delete variables that we actually still want. Those variables, which are still needed, are the only state that Islandora Google Scholar leaves behind. There's no DB tables to clean up, which is the usual reason why you'd want to uninstall something. Thoughts?

bondjimbond:
@jonathangreen Uninstalling Islandora Google Scholar would remove these variables:

```
function islandora_google_scholar_uninstall() {
  $variables = array(
    'islandora_scholar_google_scholar_search_enabled',
    'islandora_scholar_google_scholar_primary_search_xpath',
    'islandora_scholar_google_scholar_default_search_xpath',
  );
  array_walk($variables, 'variable_del');
}
```

Since they're all correctly named for Islandora Scholar (and not Islandora Google Scholar), and we want to keep those variables, I think we're happily safe from having to uninstall anything.

bondjimbond:
@bryjbrown The "missing module" error does not go away after running update.php, even after clearing the cache.

jonathangreen:
My thinking for uninstalling is that there would still be an entry in the Drupal system table for the module and Drupal would complain about it if we don't do something about it. I think we would be safe to delete this entry by hand, which would be roughly the same as uninstalling the module without running hook uninstall. I believe this is stored in the `system` table (and possibly cached).

islandora_scholar.install
```diff
@@ -149,3 +149,15 @@ function islandora_scholar_update_7103(&$sandbox) {
   $t = get_t();
```
bondjimbond, line 150: Don't we need to move the variables from islandora_google_scholar.install into islandora_scholar.install?

bryjbrown:
@bondjimbond Was Islandora Google Scholar currently enabled? I *think* the `module_exists()` function I used only triggers if the module is both installed AND enabled, so if IGS was installed but disabled, it cause weirdness?

islandora_scholar.install
bryjbrown: I believe they are already there.

bondjimbond:
@bryjbrown Was disabled when I did it, as far as I know. I'm gonna check out the main branch, enable, disable, and checkout again. Will report back in a moment..

jonathangreen:
I think we need to do roughly what's outlined [here](https://www.drupal.org/node/2487215?page=1).

```php
/**
 * Delete {system} records for long-lost modules.
 */
function MYMODULE_update_7100() {
  $modules = array(
    'old_module1',
    'old_module2',
    'old_module3',
  );
  db_delete('system')
    ->condition('name', $modules, 'IN')
    ->condition('type', 'module')
    ->execute();
}
```

bondjimbond:
Confirmed: disabled module, checked out new branch... Missing module message still appears. I think @jonathangreen's solution above should do the trick.

bryjbrown:
@jonathangreen @bondjimbond Alrighty, y'all were right, theres still an entry for IGS in the Drupal system table just like @jonathangreen said. I've added the suggested code from @jonathangreen to the update hook to clear it.

bryjbrown:
Just rebuilt the machine and ran update.php, which deletes IGS's entry from the system table on my machine.

islandora_scholar.install
bondjimbond: Egad, you're right! They're in both files:

https://github.com/Islandora/islandora_scholar/blob/7.x/islandora_scholar.install#L45-L47
https://github.com/Islandora/islandora_scholar/blob/7.x/modules/islandora_google_scholar/islandora_google_scholar.install#L13-L15

That's... That's bad, right?

bondjimbond (approved):
It works! Good change.

bondjimbond:
@jonathangreen Please feel free to hit the merge button after the 24 hours have passed; my vacation starts tomorrow and I won't be Githubbing anytime soon.

jonathangreen:
I'd like to ping @Islandora/7-x-1-x-committers and specifically @adam-vessey and see if there is any feedback with how we are doing the submodule uninstall by directly manipulating the system table. I'm pretty happy with the solution, but some more eyes 👀 would be good in case someone sees something else we missed. 

I'll give this one a couple days next week for feedback before merging.

adam-vessey:
@jonathangreen: Generally, I'm uncomfortable with the amount of large changes being made to modules without deprecation: What if people out in the wild have dependencies on `islandora_google_scholar`, for whatever reason? Once modules are released to the wild, there's no real way to know how they might be used... as such, straight up deleting/coalescing into another modules seems... imprudent?

... Also, given that this is essentially all that the `islandora_google_scholar` module does, it seems to imply that it was desirable at some point to have the inclusion of these tags be optional, leaving it up to those running sites to decide what modules to install... especially given other modules such as [`metatag`](https://www.drupal.org/project/metatag/), which might be used to similar effects, and more... (not sure about [`metatag_google_scholar`](https://www.drupal.org/project/metatag_google_scholar), but... it's a thing that exists).

In an install that was using `metatag` (or any other means to set their own meta-tags), could this change lead to the duplication of tags, possibly with different contents? Would this duplication be an issue? Shrug?

bryjbrown:
@adam-vessey I don't think `metatag` or `metatag_google_scholar` would work for Islandora objects since they aren't Drupal nodes.

I also can't imagine what modules might depend on IGS since it doesn't have an API file, but I suppose its possible and you are right that we can't really know.

Would adding a toggle (enabled by default) to the Islandora Scholar admin menu that displays/supresses metatags make you feel more comfortable with this?

jonathangreen:
@adam-vessey Thank you for reviewing, I always appreciate your feedback. I do get your point, but I think in this instance this merge is warranted. The folks from Google Scholar specifically requested that we be providing metatags as part of the scholar module by default, as they are essentially required for google scholar indexing, and they saw enough Islandora sites without them to reach out and ask us to make them enabled by default. 

Although it is possible someone is depending on this module somewhere. I think depending on a submodule of a project, then updating that project blindly, is somewhat risky behaviour on the implementers side. Is there any specific DGI modules or client modules you are worried about? In that case I would be more concerned than if its just a general worry.

If we don't think that everyone wants metatags (which I am fairly unconvinced of), I think a toggle may be the best thing, to allow them to be tuned off. My thoughts would be to wait until someone needs this toggle and then add it, but I would be open to the toggle if it helps with consensus on this ticket.

willtp87:
I think that this change can happen immediately and the transition be smoothed by a deprecation cycle rather easily.  It would just require leaving islandora_google_scholar around with it's functions having a deprecation message and pointed at the new islandora_scholar copies. We would also want to implement the suggested toggle to handle any of the scenarios Adam has mentioned.

bryjbrown:
@jonathangreen @willtp87 @adam-vessey The toggle has been implemented in the latest commit, enabled by default.

As for keeping a copy of the IGS submodule included, that would definitely force us to change the update hook since we are currently forcibly blowing up all remnants of it from the system.

@dannylamb @DiegoPino Any thoughts on this situation?

jonathangreen:
Going to tag @whikloj since he's the release manager, see if he has feedback on this one. 

I personally don't think that removing a module like this should warrant a deprecation cycle for the reasons I stated above, but I'll leave that up to the committers consensus.

jonathangreen (approved):
This looks good from my perspective 🎸🤘🏻

Also thanks @bryjbrown for doing all the work to make the good feedback happen.

PR #333, opened 2020-08-21
Alters the code to delete accessCondition and physicalDescription at file upload

bondjimbond:
**JIRA Ticket**: (https://jira.lyrasis.org/browse/ISLANDORA-2515)

Tweaks the code to delete accessCondition and physicalDescription at file upload and makes it a bit more precise.

# What's new?

Previously, any metadata in the physicalDescription or accessCondition elements would be deliberately removed at the PDF Upload stage. I'm guessing the idea was to "make room" for the "document version" and "permission" selections, which are written to those fields, but there's no reason you can't have both. 

Now the existing elements remain, and the new elements are simply appended. They can coexist in harmony.

# How should this be tested?

1. Ingest an object as normal. Include some metadata in an accessCondition and/or physical Description element.
2. Include a PDF, select some options under the permission and version section.
3. Review metadata - your submitted accessCondition and physicalDescription are gone.
4. Check out this branch.
5. Ingest a new object (not "edit metadata"), do the same thing.
6. Review metadata - now both the submitted fields and the automatically-written fields should be present.

# Interested parties
@bryjbrown @DonRichards 
@Islandora/7-x-1-x-committers

### DIFF
--- a/includes/upload.tab.inc
+++ b/includes/upload.tab.inc
@@ -223,8 +223,15 @@ function islandora_scholar_add_usage_and_version_elements_to_mods(AbstractObject
   $xpath->registerNamespace('mods', $namespace);
 
   // Remove earlier instances of mods:accessCondition and mos:physicalDescription created by this form.
-  $access_condition_query = '//mods:accessCondition[@type="use and reproduction"][text()="author"] | 
+  $permissions_config = explode("\n", variable_get('islandora_scholar_use_permissions', 'Publisher\nAuthor'));
+  $access_condition_query = '//mods:accessCondition[@type="use and reproduction"][text()="author"] |
     //mods:accessCondition[@type="use and reproduction"][text()="publisher"]';
+  foreach ($permissions_config AS $permission) {
+    $access_condition_query = $access_condition_query . ' | //mods:accessCondition[@type="use and reproduction"][text()="' . $permission . '"]';
+  }
+/*
+  $access_condition_query = '//mods:accessCondition[@type="use and reproduction"][text()="author"] | 
+    //mods:accessCondition[@type="use and reproduction"][text()="publisher"]'; */
   $physical_description_query = '//mods:physicalDescription[@authority="local"]';
   $results = $xpath->query("{$access_condition_query} | {$physical_description_query}");
   foreach ($results as $result) {
--- a/includes/upload.tab.inc
+++ b/includes/upload.tab.inc
@@ -227,11 +227,9 @@ function islandora_scholar_add_usage_and_version_elements_to_mods(AbstractObject
   $access_condition_query = '//mods:accessCondition[@type="use and reproduction"][text()="author"] |
     //mods:accessCondition[@type="use and reproduction"][text()="publisher"]';
   foreach ($permissions_config AS $permission) {
+    $permission = str_replace( "\r", "", $permission);
     $access_condition_query = $access_condition_query . ' | //mods:accessCondition[@type="use and reproduction"][text()="' . $permission . '"]';
   }
-/*
-  $access_condition_query = '//mods:accessCondition[@type="use and reproduction"][text()="author"] | 
-    //mods:accessCondition[@type="use and reproduction"][text()="publisher"]'; */
   $physical_description_query = '//mods:physicalDescription[@authority="local"]';
   $results = $xpath->query("{$access_condition_query} | {$physical_description_query}");
   foreach ($results as $result) {
--- a/includes/upload.tab.inc
+++ b/includes/upload.tab.inc
@@ -221,14 +221,6 @@ function islandora_scholar_add_usage_and_version_elements_to_mods(AbstractObject
   $doc->loadXML($object['MODS']->content);
   $xpath = new DOMXPath($doc);
   $xpath->registerNamespace('mods', $namespace);
-  // Remove all instances of mods:accessCondition and mods:physicalDescription
-  // from every mods:mods element instance.
-  $access_condition_query = '//mods:accessCondition[@type="use and reproduction"]';
-  $physical_description_query = '//mods:physicalDescription[@authority="local"]';
-  $results = $xpath->query("{$access_condition_query} | {$physical_description_query}");
-  foreach ($results as $result) {
-    $result->parentNode->removeChild($result);
-  }
   // Regardless of the number of mods:mods elements in the document we only
   // add the usage and version to the first one.
   $results = $xpath->query('//mods:mods[1]');
--- a/includes/upload.tab.inc
+++ b/includes/upload.tab.inc
@@ -221,6 +221,16 @@ function islandora_scholar_add_usage_and_version_elements_to_mods(AbstractObject
   $doc->loadXML($object['MODS']->content);
   $xpath = new DOMXPath($doc);
   $xpath->registerNamespace('mods', $namespace);
+
+  // Remove earlier instances of mods:accessCondition and mos:physicalDescription created by this form.
+  $access_condition_query = '//mods:accessCondition[@type="use and reproduction"][text()="author"]';
+  $physical_description_query = '//mods:physicalDescription[@authority="local"]';
+  $results = $xpath->query("{$access_condition_query} | {$physical_description_query}");
+  foreach ($results as $result) {
+    $result->parentNode->removeChild($result);
+  }
+
+
   // Regardless of the number of mods:mods elements in the document we only
   // add the usage and version to the first one.
   $results = $xpath->query('//mods:mods[1]');
--- a/includes/upload.tab.inc
+++ b/includes/upload.tab.inc
@@ -223,7 +223,8 @@ function islandora_scholar_add_usage_and_version_elements_to_mods(AbstractObject
   $xpath->registerNamespace('mods', $namespace);
 
   // Remove earlier instances of mods:accessCondition and mos:physicalDescription created by this form.
-  $access_condition_query = '//mods:accessCondition[@type="use and reproduction"][text()="author"]';
+  $access_condition_query = '//mods:accessCondition[@type="use and reproduction"][text()="author"] | 
+    //mods:accessCondition[@type="use and reproduction"][text()="publisher"]';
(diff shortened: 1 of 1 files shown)


bondjimbond:
One Travis build needs to be restarted - I'm not sure why I can't do that.

DonRichards:
@bondjimbond I'll restart it

manez:
Did travis change something? I can't find the restart buttons either

DonRichards:
They changed the wording I think "trigger a build"

adam-vessey:
This is rather a breaking change. There may be instances out there which expect there to be only a single value for each of these elements in a document... I could see introducing a toggle around them such that you could disable this functionality in specific environments, but it should default to the current behaviour...

Also, consider the edge case of removing and reading re-adding a PDF multiple times: you would end up with multiple of these statements, yes?

bondjimbond:
> There may be instances out there which expect there to be only a single value for each of these elements in a document... 

Can you describe what kind of setup would actually require single values and break when there are multiples? Would such a site actually have ingest forms adding this data only to be overwritten?

> Also, consider the edge case of removing and reading re-adding a PDF multiple times: you would end up with multiple of these statements, yes?

Possible, will have to test.

bondjimbond:
> Also, consider the edge case of removing and reading re-adding a PDF multiple times: you would end up with multiple of these statements, yes?

> Possible, will have to test.

Update: Yes, turns out this is a problem. I'm going to re-add the deleted lines and try a different approach. I think the existing `physicalDescription` line is actually OK, as it only deleted the "local" versions. The problem is the too-broad `accessCondition` approach.

Here's the big issue:

`$access_condition_query = '//mods:accessCondition[@type="use and reproduction"]';`

This query needs to be tightened up so that it only hunts for elements where the text is "author" OR "publisher".

Hints on that would be appreciated...

bondjimbond:
The latest commits solve the problem of repeating values. Now:

- Checks for the existence of `accessCondition type="use and reproduction"`, but ONLY if the value is `author` or `publisher`, and deletes
- Checks for `physicalDescription authority="local"` and deletes
- Leaves all other instances alone

I think this should serve the majority of cases. I don't know how we could handle custom values for the accessCondition in the upload form - they would end up getting repeated - but that could be dealt with on an ingest form anyway.

bondjimbond:
For @bryjbrown here's some MODS for testing:
```
<mods xmlns="http://www.loc.gov/mods/v3" xmlns:mods="http://www.loc.gov/mods/v3" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xlink="http://www.w3.org/1999/xlink">
<titleInfo>
  <title>Testing the Deleted Fields</title>
  <subTitle>Test</subTitle>
</titleInfo>
<genre authority="marcgt">article</genre>
<originInfo>
  <dateIssued>1924</dateIssued>
</originInfo>
<name type="personal">
  <role>
    <roleTerm authority="marcrelator" type="text">author</roleTerm>
  </role>
  <namePart type="given">Brandon</namePart>
  <namePart type="family">Weigel</namePart>
  <affiliation>Simon Fraser University</affiliation>
</name>
<typeOfResource>text</typeOfResource>
<accessCondition type="use and reproduction" displayLabel="Rights Statement">http://rightsstatements.org/vocab/NoC-OKLR/1.0/</accessCondition>
<accessCondition type="use and reproduction">publisher</accessCondition>
<physicalDescription authority="local">PRE-PUBLICATION</physicalDescription>
</mods>
```

bryjbrown:
@bondjimbond Just tested in the vagrant VM and can confirm that this seems to work as you say: prior accessCondition and physicalDescription elements were left alone, and additional ones were added. If the document was deleted and then a new documnet is uploaded with different values, the old auto-added values are replaced with the new one. 

While I agree with @adam-vessey that there is a hidden issue of instances in the wild that assume only one of these elements will be there, I would say that those problems lie more in the assumption of one element than with this PR making multiples, and that any component that assumes only one of these elements is actually the thing that needs to be fixed since MODS allows it.

bondjimbond:
> While I agree with @adam-vessey that there is a hidden issue of instances in the wild that assume only one of these elements will be there, I would say that those problems lie more in the assumption of one element than with this PR making multiples, and that any component that assumes only one of these elements is actually the thing that needs to be fixed since MODS allows it.

I couldn't have put it better myself. We should not let the possibility that someone might have hung their repository's display on the weird assumption that metadata should always have a very specific configuration stop this fix from happening - particularly since such a repository should/would not have the automatically-deleted elements in their ingest forms to begin with.

I'm going to look at one more small tweak, then I'll ask you to approve/merge..

adam-vessey:
All I'm saying is that changing the default baked-in metadata profile is backwards breaking, as in, you can't rule out that things expecting a given structure exists. Typically, such a change would have a deprecation cycle for the old behaviour to allow time for users of the module, instead of immediately putting them into force, yes?

... could even go so far as to change up how the given state is stored to avoid overloading `accessCondition[@type='use and reproduction']` with multiple types of values... anyway...

... somewhat of a moot point, given that this is no longer really _changing_ the built-in profile, but being more specific to it (with the additional predicate in the XPath).

bondjimbond:
Final(?) changes have been pushed.

Now, the accessCondition search matches the following text values for that element:
- author
- publisher
- any values configured in a custom configuration of that field

I had to keep "author" and "publisher" hard-coded because of a weird quirk of this module: if you don't specify custom values, "author" and "publisher" (lowercase) are hard-coded defaults. If you *do* specify custom values, the defaults are "Author" and "Publisher" (uppercase initial letter).

It's very weird, but I think this is the best possible approach. Any questions for your review @bryjbrown?

DiegoPino:
@bondjimbond @adam-vessey @bryjbrown  as a non Scholar user i have a question here: I can not get my head around allowing the flexibility of MODS Islandora wide and then imposing one flavor of MODS here but then also allowing XML Forms to simply bypass any decision made by this form here. Not judging, but feels like scholar was not core and almost its own universe.

I appreciate all the code gymnastics you made @bondjimbond to address concerns and support any option-deprecation-existing practice, but i feel its a futile effort to assume Islandora can-should enforce one way of seeing MODS when there is not even core code that check if a thing is actually MODS! (or Apple Core..) or not.


So, seeing that this adds a config setting, what about instead add a checkbox to the form itself that allows old v/s new behavior to exist? And that Checkbox can be an admin driven permission (administer metadata?) triggered by (like shown only if the situation of a non conforming to this need MODS is already in place). So checkbox checked, you get your old stuff and illusion (sorry, it is an illusion) of single metadata profile based on MODS, unchecked, what @bondjimbond needs? All this because this could be a little bit of over engineering that would need also documentation. 

Just my 25 cents here, i'm not emotionally involved in this in any way. No need to respond is this is again me being silly old person

bondjimbond:
@DiegoPino No config setting is added here, I'm just working with the intricacies of the existing config setting that isn't playing well with the rest of its code. I think the code I've added here does the job and performs the behaviour users expect at this point without conflicting with existing users' work -- I'd be really reluctant to take another approach.

bondjimbond:
>  as a non Scholar user i have a question here: I can not get my head around allowing the flexibility of MODS Islandora wide and then imposing one flavor of MODS here but then also allowing XML Forms to simply bypass any decision made by this form here. Not judging, but feels like scholar was not core and almost its own universe.

@DiegoPino Yes, Scholar is a very messy module, and because it's been in use so long, everything we do with it is a patch instead of a real solution. My feeling with this PR is that this is the best we can do with what we've got.

DiegoPino:
All good @bondjimbond and thanks for replying. I get it. Sometimes there is only so much you can do. Thanks a lot

bondjimbond:
Additional testing instructions, since I made those tweaks to the accessCondition search code:
- Ingest your object as above
- In Scholar config, check the "Specify use permissions" box and add a custom entry
- On your object, on the Document tab, replace the document and change the permissions
- Review the MODS at datastream/MODS/view and confirm that the old metadata is gone, new metadata is present, and the metadata you provided (not provided by the module) is still there

bryjbrown:
@bondjimbond So I just did a test with an object that had no PDF and a MODS file that started with these values:
- `<accessCondition type="use and reproduction">Don't touch this</accessCondition>`
- `<physicalDescription authority="local">Don't touch this either</physicalDescription>`

When I uploaded a document and selected custom my custom values for Document Version and Use Permission, the MODS record got an additional `<accessCondition type="use and reproduction">` with the custom value (the original accessCondition remained) but the original `<physicalDescription authority="local">` was overwritten with the old value being lost and replaced with the new one.

TBH I don't understand the changes that have been introduced since the last time I tested so you can determine if this is intended behavior or not.

bondjimbond:
@bryjbrown Yes, this is the intended behaviour. `<physicalDescription authority="local">` is not a normal MODS thing to do, so given that the module is the only thing that writes this element in normal circumstances, the module remains allowed to replace it.

bondjimbond:
> TBH I don't understand the changes that have been introduced since the last time I tested so you can determine if this is intended behavior or not.

The changes are, basically, instead of disabling the "delete" functionality, it preserves it, but limits the deleting to the elements that are created by Scholar.

bryjbrown (approved):

bryjbrown:
Just reviewed the code change and it looks good and matches up with the behavior you outlined @bondjimbond. I'd have a strong preference with @DonRichards taking a look at this too since XPaths are his wheelhouse, but I'm approving the PR regardless.

bondjimbond:
Huzzah! Thanks @bryjbrown - it's an esoteric little change for some, but it's a big deal for our sites.

bryjbrown:
Ping @DonRichards

bondjimbond:
If @DonRichards doesn't have anything to say, can we merge?

DiegoPino:
@bondjimbond yes, was approved 11 days ago so yeah, some needs to merge this.Priority is on everyone involved in the review and testing but will press the green button in 24 hours if nobody does before.

bondjimbond:
@DiegoPino Sounds like there are no objections, so if you'd like to press the button I'd appreciate it!

